PR #5, opened 2018-10-10
Initial setup

bitPogo:


bitPogo:
- Do we want member ordering?

wiese:
@bitPogo Mind explaining "member ordering"? Is that alphabetical ordering of object members you want to see in liniting?

jakobw:
@wiese I think this is about https://palantir.github.io/tslint/rules/member-ordering/. I'm not the biggest fan, because I think it's nice to put e.g. a private method right below the public method that uses it, but I don't have strong feelings about it.

bitPogo:
@jakobw +1

wiese:
@jakobw Thanks for the pointer. So yes, enforcing of member ordering (as it reads on the can). With its many config options it's not out of the question that the order you fancy is achievable with it, so why not use it if it eliminates room for endless discussions. But could/should totally be a separate step instead of holding this back.
Will not vote as I did not try yet.

wiese:
@jakobw Thanks for the pointer. So yes, enforcing of member ordering (as it reads on the can). With its many config options it's not out of the question that the order you fancy is achievable with it, so why not use it if it eliminates room for endless discussions. But could/should totally be a separate step instead of holding this back.
Will not vote as I did not try yet.

PR #18, opened 2018-10-23
[WIP] Term Wrapper

bitPogo:


public/index.html
```diff
@@ -6,7 +6,10 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title>wikibase-termbox</title>
-	<script> var mw; </script>
+	<script> 
+		var mw; 
+		var langKey = 'de';
```
jakobw: I hope this won't be needed

src/client-entry.ts
```diff
@@ -4,6 +4,11 @@ import store from '@/store';
 
 Vue.config.productionTip = false;
 
+declare const mw: any;
+
+store.state.languageKey = mw.uls.getBrowserLanguage();
```
jakobw: This should be fine.

src/client-entry.ts
```diff
@@ -4,6 +4,11 @@ import store from '@/store';
...
+store.state.termData = mw.config.get( 'wbEntity' );
```
jakobw: Only labels, descriptions and aliases are needed in the store, so not the whole blob should be added to the store, especially since we will likely only send the needed data (again, only labels, descriptions, aliases) with the server side request.

If we do a `JSON.parse( mw.config.get( 'wbEntity' ) )` then we can separately add `id`, `labels`, `descriptions`, and `aliases` to the store. This could even be extracted into a small class that gets the blob as an input and fills the store, so it's easier to test.

src/components/lib/exceptions/LogicalException.ts
```diff
@@ -0,0 +1,5 @@
+export default class LogicalException {
+	constructor( ...Parameter: any[] ) {
+		console.trace( new Error( ...Parameter ) );
```
bitPogo: See here for stacktrace

src/components/lib/exceptions/TypeException.ts
```diff
@@ -0,0 +1,5 @@
+export default class TypeExcpetion {
+	constructor( ...Parameter: any[] ) {
+		console.trace( new TypeError( ...Parameter ) );
```
bitPogo: See here for stacktrace

bitPogo:
Done with #26 an #30

PR #21, opened 2018-10-31
Introduce store as place to be

bitPogo:


bitPogo:
No clue, why the tests still failing, even after wikibase.termbox.js check-in.

wiese:
We only introduced the dist comparison yesterday, so it may have problem - however the [CI log](https://travis-ci.org/wmde/wikibase-termbox/builds/449761310?utm_source=github_status&utm_medium=notification) shows a couple of errors (unsure why it does not turn red at that point already). Maybe you can look into these first?

bitPogo:
The console trace is gone now and it provides still a nicely stacktrace.

wiese:
I can reproduce the diff when locally rebuilding the branch. Did you use the container-based approach to run npm when building it?

bitPogo:
Nope

wiese:
@bitPogo Then please, until we have a more stable way to ensure code-dist-consistency, go with the way to build that is mentioned in the README. That's not a blocker for us starting to look at it, however. Mind you, we do not really work on this project today so it'll have to wait until next week.

wiese:
Replaced by #26

PR #23, opened 2018-11-01
travis: run in container & check dist/ build matches

wiese:
* Ensure the dist folder matches the rest of the repo to avoid e.g. #20 
* run npm scripts in container we use in dev, too

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,16 @@
 language: node_js
 
-node_js:
-  - '8'
+services:
+  - docker
+
+install:
+  - cp .env.example .env
+  - docker-compose run --rm node npm ci
 
 script:
-  - npm run build-server
-  - npm run lint
-  - npm run test
+  - docker-compose run --rm node npm run build-server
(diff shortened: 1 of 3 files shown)


wiese:
A violating build would look like this: https://travis-ci.org/wmde/wikibase-termbox/builds/449284657

.travis.yml
```diff
@@ -1,13 +1,20 @@
...
 script:
-  - npm run build-server
-  - npm run lint
-  - npm run test
+  - docker-compose run --rm node npm run build-server
+  - docker-compose run --rm node npm run lint
+  - docker-compose run --rm node npm run test
+  - docker-compose run --rm node npm run build
+  # check that dist contains the latest build result
+  - git status && git diff-index --quiet HEAD dist/
```
jakobw: Cool! TIL
```
       --quiet
           Disable all output of the program. Implies --exit-code.
       --exit-code
           Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.
```

PR #26, opened 2018-11-05
Introduce the store as place to be

bitPogo:
The store should be the main place for several objects, which are needed
by different components of the frontend with a consitent set of data.
The store itself is based on Vuex and differs in several modules,
which are also encapsulated by their namespaces. Methods (eg.
mutations, get or actions), which are based on tow or more different
modules, are assembled at the root of the store.

One module itself is crumbled also in several parts:
1. Properties, which provides a interface with the properies and their
types of the module,
2. Mutations, which define the major accesspoint for the module,
3. Getters, which are accesspoint to get values of the module,
4. Module, which define the module itself,
5. MethodNames, which define the method names of the module,
6. exceptions, which contains several exceptions which are specific for
the module,
7. Helper and other utlity classes of the module.

To provide this exception based approach this patch also introduce
several base exception types, which are stored in src/common/exceptions.
The common directory is supposed to be the place, where classes,
interfaces, etc are deposited.

So far the store contains 2 objects -> language and entity.
Entity contains a id, type. Also it contains several labels, descriptions and aliases,
which are refer to language key ( eg. en or de ). To get a specific
label of a language or something similar, the entity-module has a get[SOMETHING]ByLanguage method.
To ensure the correct type for the module you have to use the EntityStripper or a derivative of it.

Language is very similar constructed to Entity.

Bug: T207150


src/store/StoreBundle.ts
```diff
@@ -0,0 +1,19 @@
...
+
+Vue.use( Vuex );
+
+const storeBundle: StoreOptions<StateInterface> = {
+	getters,
+	modules: {
+		entity,
+		language,
+	},
+	strict: true,
```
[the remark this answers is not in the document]
wiese:
```suggestion
	strict: process.env.NODE_ENV !== 'production',
```

bitPogo:
I do not agree to avoid namespaces. Now, when you look at the store, you'll notice that methods are appended to root at runtime. In so far, the store represents only one object (in the surface view) and does not encapsulate several objects that live in the store.

src/common/exceptions/ParameterException.ts
```diff
@@ -0,0 +1,8 @@
+/* tslint:disable: variable-name */
```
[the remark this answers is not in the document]
bitPogo: Done

src/mock-data/MwUls.ts
```diff
@@ -0,0 +1,10 @@
+import * as entity from './data/Q64_data.json';
+
+export default class MWConfig {
```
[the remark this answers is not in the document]
bitPogo: Done

src/store/root/Getters.ts
```diff
@@ -0,0 +1,34 @@
...
+	ENTITY_GET_LABEL_BY_LANGUAGE,
+	ENTITY_GET_ALIASES_BY_LANGUAGE,
+	ENTITY_GET_DESCRIPTION_BY_LANGUAGE,
+} from '@/store/Entity/EntityMethodNames';
+import {
+	NS_ENTITY,
+	NS_LANGUAGE,
+} from '@/store/root/Namespaces';
+
+export const rootGetters: GetterTree<StateInterface, any> = {
```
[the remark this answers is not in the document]
bitPogo: done

src/store/Entity/EntityMethodNames.ts
```diff
@@ -0,0 +1,11 @@
+/* Mutations */
```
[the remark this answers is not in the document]
bitPogo: Outdated

src/store/StoreBundle.ts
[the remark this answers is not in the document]
bitPogo: Done

src/store/StoreBundle.ts
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export const store = new Vuex.Store<StateInterface>( storeBundle );
```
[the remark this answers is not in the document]
bitPogo: Done

src/store/Entity/EntityModule.ts
```diff
@@ -0,0 +1,27 @@
+import { Module } from 'vuex';
+import EntityProperties from '@/store/Entity/EntityProperties';
+import { getters } from '@/store/Entity/EntityGetters';
+import { mutations } from '@/store/Entity/EntityMutations';
+import StateInterface from '@/store/root/StateInterface';
+
+const entityInit: EntityProperties = {
```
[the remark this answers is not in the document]
bitPogo: Outdated

jakobw:
For the lack of a better place to document this, here is what I found out about module namespaces and `mapGetters` magic.

We can have `namespaced: true` for all our store modules. Getters within those modules will be registered with the funky `[module]/[getter]` notation. We can then do something like the following whenever the getters are used in a component:
```
// maps the entity/id getter to an entityId computed method

@Component( {
	computed: {
		...mapGetters( 'entity', {
			entityId: 'id',
		} ),
	},
} )
```

src/store/Entity/EntityStripper.ts
```diff
@@ -0,0 +1,101 @@
+import Dictionary from '@/common/interfaces/Dictionary';
```
jakobw: Wikibase has a service called `EntityInitializer` which does basically the same thing. Conceptually, I find it odd that the `ENTITY_INIT` action takes a service emitting different parts of the entity instead of the entity object itself.

Here is my yes-model:
* rename this to `EntityInitializer`
* have a single method `getEntityFromJson( serializedEntity: string ): Entity` - `Entity` being an interface containing all the required fields
* pass the Entity object instead of the initializer into the action

src/store/Entity/EntityStripper.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	private getNestedLabels(): Dictionary<string> {
+		return this.getNestedValues( this.unstrippedEntity, 'labels' );
+	}
+
+	private getNesatedDescriptions(): Dictionary<string> {
+		return this.getNestedValues( this.unstrippedEntity, 'descriptions' );
+	}
+
+	private checkPrerequisites(): void {
```
jakobw: IMO this is a bit overly cautious and could get away with checking that `type` and `id` exist on the given JSON object. The rest should ideally be taken care of by strict typing.

src/store/Entity/Getters.ts
```diff
@@ -0,0 +1,47 @@
+import { GetterTree } from 'vuex';
+import Dictionary from '@/common/interfaces/Dictionary';
+import StateInterface from '@/store/Entity/Properties';
+
+export const getters: GetterTree<StateInterface, any> = {
```
jakobw: getters without (!) additional parameters are accessed like object properties, so IMO they should also be named as such e.g. `id` instead of `getId`, `labels` instead of `getLabels`. `getLabelByLanguage` is fine, though.

src/store/Entity/Module.ts
```diff
@@ -0,0 +1,21 @@
+import { Module } from 'vuex';
```
jakobw: This file is the entry point to the entity module in a way. The common convention is to call it `index.ts`.

src/store/Entity/Mutation.Types.ts
```diff
@@ -0,0 +1 @@
+export const ENTITY_INIT: string = 'initilizeEntity';
```
jakobw: The conventional name is `mutationTypes.ts`

src/store/Entity/Mutations.ts
```diff
@@ -0,0 +1,18 @@
+import { MutationTree } from 'vuex';
```
jakobw: The conventional name is `mutations.ts` (lowercase)

src/store/Entity/Properties.ts
```diff
@@ -0,0 +1,9 @@
+import Dictionary from '@/common/interfaces/Dictionary';
```
jakobw: `Property` is a pretty loaded term in the Wikibase world. The convention for the state interface name that I've seen is `[ModuleName]State`, so this should be `EntityState.ts`

src/store/Entity/exceptions/InvalidEntityException.ts
```diff
@@ -0,0 +1,8 @@
+/* tslint:disable: variable-name */
```
jakobw: This comment is not needed.

src/store/Entity/exceptions/InvalidEntityStripperException.ts
```diff
@@ -0,0 +1,7 @@
+import ParameterException from '@/common/exceptions/ParameterException';
```
jakobw: This file is not needed, I think

src/store/User/Getters.ts
```diff
@@ -0,0 +1,8 @@
+import { GetterTree } from 'vuex';
+import Properties from '@/store/User/Properties';
+
+export const getters: GetterTree<Properties, any> = {
+	getPrimaryLanguage( state: Properties ) {
```
jakobw: `primaryLanguage`

src/store/User/Module.ts
```diff
@@ -0,0 +1,17 @@
+import { Module } from 'vuex';
```
jakobw: `index.ts`

src/store/User/Mutation.Types.ts
```diff
@@ -0,0 +1 @@
+export const LANGUAGE_INIT: string = 'initilizeLanguage';
```
jakobw: `mutationtypes.ts`

src/store/User/Mutations.ts
```diff
@@ -0,0 +1,16 @@
+import { MutationTree } from 'vuex';
```
jakobw: `mutations.ts`

src/store/User/Properties.ts
```diff
@@ -0,0 +1,3 @@
+export default interface UserProperties {
```
jakobw: `UserState.ts`

jakobw:
Re our recent discussion: FWIW here's a running example for a unit test for the current implementation of User/Getters.js:

```js
import { getters } from '@/store/User/Getters';

describe( 'User/getters', () => {
	it( 'returns the primary language key', () => {
		expect( getters.getPrimaryLanguage( { PrimaryLanguage: 'en' }, null, null, null ) )
			.toMatch( 'en' );
	} );
} );
```

Whether this is better than testing the integration of store + getters + getter constants is indeed debatable.

bitPogo:
Is there any reason, why some files are only accessible by root?

PR #27, opened 2018-11-06
tsconfig: check for unused imports

wiese:
See https://www.typescriptlang.org/docs/handbook/compiler-options.html

This now run during the build steps instead of the lint step.

### DIFF
--- a/src/mockup-entry.ts
+++ b/src/mockup-entry.ts
@@ -1,7 +1,9 @@
 import MWConfig from '@/mock-data/MwConfig';
 
-declare let mw: any;
+declare global {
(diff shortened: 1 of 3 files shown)


wiese:
Yields things like

```
 ERROR  Failed to compile with 1 errors                                                                                                                                                             16:40:02

 error  in /app/src/server-entry.ts

ERROR in /app/src/server-entry.ts
3:1 'store' is declared but its value is never read.
    1 | import Vue from 'vue';
    2 | import App from '@/components/App.vue';
  > 3 | import store from '@/store';
      | ^

PR #29, opened 2018-11-07
container node: ensure right user

wiese:
The node docker image comes with a 'node' user that can (and should) be
used instead of root. Setting its user id to the one of your user on the
host system ensures that files generated into the mounted volumes are
owned by you.
Make sure to give the README another read!

See https://github.com/nodejs/docker-node/blob/526c6e618300bdda0da4b3159df682cae83e14aa/8/jessie/Dockerfile#L3

Interestingly, even though we've been running npm through the container before, I now see changes to `package-lock.json` after running `npm install` (`"optional": true` stuff) - not sure yet, if we should check that in or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ services:
 
 install:
   - cp .env.example .env
+  - docker-compose build --build-arg UID=$(id -u) --build-arg GID=$(id -g) node
   - docker-compose run --rm node npm ci
 
 script:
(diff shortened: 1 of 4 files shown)


wiese:
@bitPogo If you used the previous setup you may find files on your disk that are owned by root. This change is to stop it from happening again. Recommendation: delete these files (may need to use `sudo`) and restore them from git. If you follow the (new) README, all project files should be created with your user in the future.

PR #31, opened 2018-11-08
Improve separation between the entity state interface and domain objects

jakobw:
* gets rid of one of the two identical `Entity` interfaces
  * objects referring to domain objects now use `FingerprintableEntity`
  * the state object uses `Entity`
* fixes a bug in the description and alias getters discovered in test refactoring
* test refactoring (more unit-y)

(Sorry, these changes are a bit all over the place. I can split it up if necessary.)


jakobw:
@wiese and I decided that it makes sense to change the store's data structures to be be closer to the wikibase json format, so this PR in itself doesn't make a lot of sense.

PR #37, opened 2018-11-12
Merge tsconfigs

bitPogo:
This patch moves all tsconfigs to the root level and set up a base file
(tsconfig.base.json), which the server (tsconfig.server.json) and client (tsconfig.json) are extending for there specific purposes.


bitPogo:
Maybe it's a good idea to make something like a tsconfig.all.json, for "npm run lint"...so we can use the vue-cli thinging for validation

jakobw:
Since the test files are not included in the `includes`, `npm run lint` no longer checks those, which is bad. I don't like having 4 `tsconfig.*.json` files, but currently don't see a way around having the `tsconfig.all.json`.

I verified that configs from the base config are inherited, so that works! :tada:

package.json
```diff
@@ -4,23 +4,25 @@
...
+    "dev-server": "concurrently \"tsc -w --project tsconfig.server.json\" \"nodemon serverBuild/server.js\"",
     "build": "NODE_ENV=production vue-cli-service build",
-    "build-server": "WEBPACK_TARGET=node vue-cli-service build && tsc --p src/server",
-    "lint": "vue-cli-service lint --no-fix",
-    "fix": "vue-cli-service lint",
-    "test": "vue-cli-service test:unit"
+    "build-server": "npm run copyClient; WEBPACK_TARGET=node vue-cli-service build",
+    "lint": "tslint --format codeFrame --project tsconfig.all.json -c tslint.json",
+    "fix": "tslint --format codeFrame --fix --project tsconfig.all.json -c tslint.json",
+    "test": "vue-cli-service test:unit;"
```
jakobw:
```suggestion
    "test": "vue-cli-service test:unit"
```

package.json
```diff
@@ -4,23 +4,25 @@
...
   "scripts": {
     "serve": "WEBPACK_TARGET=dev vue-cli-service serve",
-    "dev-server": "concurrently \"tsc -w --p src/server\" \"nodemon serverBuild/server.js\"",
+    "dev-server": "concurrently \"tsc -w --project tsconfig.server.json\" \"nodemon serverBuild/server.js\"",
     "build": "NODE_ENV=production vue-cli-service build",
-    "build-server": "WEBPACK_TARGET=node vue-cli-service build && tsc --p src/server",
-    "lint": "vue-cli-service lint --no-fix",
-    "fix": "vue-cli-service lint",
-    "test": "vue-cli-service test:unit"
+    "build-server": "npm run copyClient; WEBPACK_TARGET=node vue-cli-service build",
```
jakobw:
```suggestion
    "build-server": "WEBPACK_TARGET=node vue-cli-service build",
```

package.json
```diff
@@ -4,23 +4,25 @@
   "private": true,
   "scripts": {
     "serve": "WEBPACK_TARGET=dev vue-cli-service serve",
-    "dev-server": "concurrently \"tsc -w --p src/server\" \"nodemon serverBuild/server.js\"",
+    "dev-server": "concurrently \"tsc -w --project tsconfig.server.json\" \"nodemon serverBuild/server.js\"",
```
jakobw: the `servers.js` path changed
```suggestion
    "dev-server": "concurrently \"tsc -w --project tsconfig.server.json\" \"nodemon serverBuild/server/server.js\"",
```

I also get a `server is running at port undefined` but I don't know if that's just me

package.json
```diff
@@ -4,23 +4,25 @@
...
-    "test": "vue-cli-service test:unit"
+    "build-server": "npm run copyClient; WEBPACK_TARGET=node vue-cli-service build",
+    "lint": "tslint --format codeFrame --project tsconfig.all.json -c tslint.json",
+    "fix": "tslint --format codeFrame --fix --project tsconfig.all.json -c tslint.json",
+    "test": "vue-cli-service test:unit;"
   },
   "dependencies": {
     "dotenv": "^6.1.0",
     "express": "^4.16.4",
+    "only": "0.0.2",
```
jakobw:
```suggestion
```

package.json
```diff
@@ -4,23 +4,25 @@
...
     "vue": "^2.5.17",
     "vue-class-component": "^6.0.0",
     "vue-property-decorator": "^7.0.0",
     "vue-server-renderer": "^2.5.17",
     "vuex": "^3.0.1"
   },
   "devDependencies": {
+	"@types/node": "^10.12.6",
```
jakobw:
```suggestion
    "@types/node": "^10.12.6",
```

tsconfig.all.json
```diff
@@ -0,0 +1,26 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "esnext",
+    "sourceMap": true,
+    "lib": [
+      "esnext",
+	  "es5",
```
jakobw:
```suggestion
      "es5",
```

tsconfig.server.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "serverBuild",
+    "target": "es5",
+    "module": "commonjs",
+	"sourceMap": true,
```
jakobw: whitespace

src/types/client/mediawiki/index.d.ts
```diff
@@ -0,0 +1,12 @@
+declare module "mediawiki" {
```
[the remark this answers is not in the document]
bitPogo: I agree, but unfortunately the renamed file does not work out with docker (for me locally). Vanilla version seems to handle it in a expected way.

src/types/client/mediawiki/index.d.ts
```diff
@@ -0,0 +1,12 @@
...
+	export interface ImmediatelyInvokingEntityLoadedHookHandler {
```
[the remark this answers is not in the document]
bitPogo: The lack of this definition causes a error in the other parts (like src/client/init.ts). The only alternative is to use 'any' at MW.hook, which I tried to avoid, because there is some sort of definition for this.

tests/unit/client/init.spec.ts
```diff
@@ -1,22 +1,28 @@
 import Vue from 'vue';
 import ImmediatelyInvokingEntityLoadedHookHandler from '@/mock-data/ImmediatelyInvokingEntityLoadedHookHandler';
 import init from '@/client/init';
+/// <reference path="src/types/client/mediawiki/index.d.ts"/>
```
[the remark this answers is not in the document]
bitPogo: I took it from here: https://www.typescriptlang.org/docs/handbook/modules.html
But it seems to work without the entire statement, so I will remove it.

bitPogo:
The problems are rooted in the fact, that we have a tsconfig that is not the configuration for the server. Alternatively we can use tsconfig.all for dev and for build we have to rename the client version in tsconfig ... and so on. However, we have already skipped this approach.

wiese:
@jakobw mentioned a hunch he had regarding using module-alias. We really need to get this out of the way, first thing tomorrow.

PR #39, opened 2018-11-12
[WIP] Unify Termbox initialization.

jakobw:


jakobw:
@wiese @bitPogo I did a thing. This is obviously unfinished, but I would appreciate some early feedback whether you think this makes sense.

My goal was to create a single place where the `App` component gets created and has its store initialized from a `TermboxRequest`. This now happens in  `buildApp.ts` (name up for discussion). I added some small piece of functionality to `App.vue` to make sure that this actually works. What you see in `server-entry.ts` is a silly workaround which should be solved by fixing our directory structure / dependency problem.

src/common/buildApp.ts
```diff
@@ -0,0 +1,20 @@
+import App from '@/components/App.vue';
+import store from '@/store';
+import TermboxRequest from '@/common/TermboxRequest';
+import { NS_ENTITY } from '@/store/namespaces';
+import { ENTITY_INIT } from '@/store/entity/mutationTypes';
+import Vue from 'vue';
+
+export default ( req: TermboxRequest ) => {
```
wiese: I think the shared TermboxRequest is a key concept in going forward (with the concrete class being to superior concept to the interface) and staying sane.

wiese:
Rebased

PR #40, opened 2018-11-13
Ensure every test has assertions

jakobw:
Without this, the following would add a passing test to our suite.

       it( 'potato', () => {
       } );


wiese:
Sample output

```
 FAIL  tests/unit/App.spec.ts
  ● App.vue › potato

    expect.hasAssertions()

    Expected at least one assertion to be called but received none.

      1 | beforeEach( () => {
    > 2 | 	expect.hasAssertions();
        | 	       ^
      3 | } );
      4 |

PR #43, opened 2018-11-15
WikibaseRepo: add way to get entity on server (api)

wiese:
Done through mwbot.
This makes our tests (ts-jest) use the tsconfig.all.json until we get to
re-organize the test folder to be able to test client-only and server-only (and possibly mixed) individually.
This is to ensure the typeRoots defined (for server) are not made obsolete by being excluded thereafter.

Bug: T207467

### DIFF
--- a/src/server/data-access/MwBotWikibaseRepo.ts
+++ b/src/server/data-access/MwBotWikibaseRepo.ts
@@ -1,0 +1,61 @@
+import WikibaseRepo from '@/server/data-access/WikibaseRepo';
+import mwbot from 'mwbot';
+import TechnicalProblem from '@/server/data-access/error/TechnicalProblem';
+import EntityNotFound from '@/server/data-access/error/EntityNotFound';
+import FingerprintableEntity from '@/datamodel/FingerprintableEntity';
+import EntityInitializer from '@/common/EntityInitializer';
+
+export default class MwBotWikibaseRepo implements WikibaseRepo {
+	private bot: mwbot;
+	private entityInitializer: EntityInitializer;
+
+	public constructor( bot: mwbot, entityInitializer: EntityInitializer ) {
+		this.bot = bot;
+		this.entityInitializer = entityInitializer;
+	}
+
+	private getEntity( id: string ): Promise<any> {
+		return new Promise( ( resolve, reject ) => {
+			this.bot.request( {
+				ids: id,
+				action: 'wbgetentities',
+			} )
+				.then( ( response: any ) => {
+					if ( !( 'entities' in response ) ) {
+						reject( new TechnicalProblem( 'wbgetentities result not well formed.' ) );
+					}
+
+					if ( !( id in response.entities ) ) {
+						reject( new EntityNotFound( 'wbgetentities result does not contain relevant entity.' ) );
+					}
+
+					const entity = response.entities[ id ];
+
+					if ( 'missing' in entity ) {
+						reject( new EntityNotFound( 'Entity flagged missing in response.' ) );
+					}
+
+					resolve( entity );
+				} )
+				.catch( ( reason: string ) => {
+					reject( new TechnicalProblem( reason ) );
+				} );
+		} );
+	}
+
+	public getFingerprintableEntity( id: string ): Promise<FingerprintableEntity> {
+		return new Promise( ( resolve, reject ) => {
+			this.getEntity( id )
+				.then( ( entity: any ) => {
+					try {
+						resolve ( this.entityInitializer.newFromSerialization( entity ) );
+					} catch ( e ) {
+						reject( new TechnicalProblem( e.message ) );
+					}
+				} )
+				.catch( ( reason ) => {
+					reject( reason );
+				} );
+		} );
+	}
+}
--- a/src/server/data-access/WikibaseRepo.ts
+++ b/src/server/data-access/WikibaseRepo.ts
@@ -1,0 +1,8 @@
+import FingerprintableEntity from '@/datamodel/FingerprintableEntity';
+
+/**
+ * Rejects to TechnicalProblem or EntityNotFound errors in case of problems
+ */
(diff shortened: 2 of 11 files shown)


src/server/data-access/MwBotWikibaseRepo.ts
```diff
@@ -0,0 +1,42 @@
+import WikibaseRepo from '@/server/data-access/WikibaseRepo';
+import mwbot from 'mwbot';
+import TechnicalProblem from '@/server/data-access/error/TechnicalProblem';
+import EntityNotFound from '@/server/data-access/error/EntityNotFound';
+import Entity from '@/store/entity/Entity';
```
jakobw: This `Entity` interface used here and in the following changes is the vuex store's entity module's state interface, which may or may not be the same as an entity object's json serialization. I was at some point very tempted to create an `EntityJson` (bad name?) or similar to have the separation, but was unsure about it.They have the same properties at the moment, but I find it a bit odd conceptually.

src/server/data-access/MwBotWikibaseRepo.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export default class MwBotWikibaseRepo implements WikibaseRepo {
+	private bot: mwbot;
+
+	public constructor( bot: mwbot ) {
+		this.bot = bot;
+	}
+
+	public getEntity( id: string ): Promise<Entity> {
```
jakobw: This might even return an actual `FingerprintableEntiy`.

src/server/data-access/WikibaseRepo.ts
```diff
@@ -0,0 +1,5 @@
+import Entity from '@/store/entity/Entity';
+
+export default interface WikibaseRepo {
```
jakobw: I like that this interface exists!

tests/unit/server/data-access/MwBotWikibaseRepo.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+			};
+			const bot = {
+				request: ( params: object ) => {
+					return Promise.resolve( results );
+				},
+			};
+
+			const api = newMwBotWikibaseRepo( bot );
+			api.getEntity( entityId ).then( ( result: Entity ) => {
+				// checking for interface type not supported, yet
```
jakobw: `FingerprintableEntity` to the rescue?

src/server/data-access/MwBotWikibaseRepo.ts
wiese: I had this first but in the end decided to take the potential failure to convert the (found) API entity into a FingerprintableEntity (happening in EntityInitializer) out of this and put it into the layer next up (to not let this class have too many responsibilities). Background being that the entity could be found by the API just fine and our EntityInitializer just lost touch with the format they come in. What do you think about that? Consequently we'd need a transport vehicle for the moment in-between (strictly speaking even if it were within MwBotWikibaseRepo).
That this uses an `Entity` interface that is part of store is indeed utter bollocks - briefly looking at the history revealed that my mental model of "Entity" (and what I assumed auto-completed) referred to what was formerly known as `@/common/interfaces/EntityType` and has been discontinued.

wiese: I now changed the interface to have a `getFingerprintableEntity()` method (instead of `getEntity()`) and have it resolve to `FingerprintableEntity`. I'm not a fan but the name should also be enough of a red flag to remind us that some further definition is due.

src/server/data-access/error/EntityNotFound.ts
```diff
@@ -0,0 +1 @@
+export default class EntityNotFound extends Error {}
```
bitPogo: Do we want that exception on the stack, like in src/common/exceptions/LogicException.ts ?

tests/unit/server/data-access/MwBotWikibaseRepo.spec.ts
```diff
@@ -0,0 +1,188 @@
+import EntityInitializer from '@/common/EntityInitializer';
+import FingerprintableEntity from '@/datamodel/FingerprintableEntity';
+import MwBotWikibaseRepo from '@/server/data-access/MwBotWikibaseRepo';
+import EntityNotFound from '@/server/data-access/error/EntityNotFound';
+import TechnicalProblem from '@/server/data-access/error/TechnicalProblem';
+import mwbot from 'mwbot';
+
+function newMwBotWikibaseRepo( bot: any, initializer?: any ) {
```
bitPogo: Do we need 'any' at bot here?

wiese: What would be the appropriate type for "random mocks I make up at will"?

src/server/data-access/error/EntityNotFound.ts
wiese: You mean generally creating only exception types that extend (our) generic exception types? I could get behind this for the sake of clustering them (less so for the stack trace part).
However I did not really have them on my radar so far and how their inheritance tree looks like. Having taken a brief look I'd shy away from building on top of them as the first class I opened shows, in my eyes, a problem: `RuntimeException` should not extend `LogicException` per my understanding. [Compare](http://php.net/manual/en/spl.exceptions.php#spl.exceptions.tree)

src/server/data-access/WikibaseRepo.ts
```diff
@@ -0,0 +1,8 @@
+import FingerprintableEntity from '@/datamodel/FingerprintableEntity';
+
+/**
+ * Rejects to TechnicalProblem or EntityNotFound errors in case of problems
```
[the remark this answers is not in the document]
wiese: Why? Should not all `WikibaseRepo`s restrict themselves to these error types. Maybe the concrete method could do with an inheritdoc, sure.

PR #44, opened 2018-11-15
Introduce module-alias

jakobw:
This allows us to use the `@` directory alias inside the generated (JS) files which otherwise would have no concept of the aliases know to the TS compiler.


src/store/entity/mutations.ts
```diff
@@ -8,8 +8,8 @@ import FingerprintableEntity from '@/datamodel/FingerprintableEntity';
 
 export const mutations: MutationTree<Entity> = {
 	[ENTITY_INIT] ( state: Entity, entity: FingerprintableEntity ): void {
-		if ( !( entity instanceof FingerprintableEntity ) ) {
-			throw new InvalidEntityException();
+		if ( !( entity.constructor.name === FingerprintableEntity.name ) ) {
```
wiese: Bit mental TODO

PR #46, opened 2018-11-16
BugFix which is for tsconfig seperation

bitPogo:


src/server/app.ts
```diff
@@ -1,17 +1,17 @@
-import fs from 'fs';
+import * as fs from 'fs';
```
jakobw: I don't think these changes are necessary.

I have the branch checked out, and this is fine: 
`git checkout master src/server/app.ts && docker-compose run --rm node npm run build-server`

bitPogo: done

jakobw: Sorry, I meant that none of the changes in `src/server/app.ts` are needed.

bitPogo: done

PR #48, opened 2018-11-16
custom errors: ensure correct prototype

wiese:
Lack of features expected from the server side (https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work)
led us to discover that the target was actually set to ES5.
This obsoletes manual prototype setting.

### DIFF
--- a/src/server/route-handler/termbox/error/EntityNotFound.ts
+++ b/src/server/route-handler/termbox/error/EntityNotFound.ts
@@ -1,6 +1,1 @@
-export default class EntityNotFound extends Error {
-	constructor( m: string ) {
-		super( m );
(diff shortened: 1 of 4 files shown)


jakobw:
Silly question, but are we sure this isn't magically taken care of by our build step? Our client `tsconfig.json` says we're compiling to `esnext` js. I would hope that after the ts -> js step, babel takes over and gets rid of browser specific issues like this?

wiese:
/me reading https://blog.sessionstack.com/how-javascript-works-the-internals-of-classes-and-inheritance-transpiling-in-babel-and-113612cdc220

wiese:
https://phabricator.wikimedia.org/T209930

PR #49, opened 2018-11-16
User language: scaffolding

jakobw:


src/components/TermBox.vue
```diff
@@ -1,21 +1,32 @@
...
+@Component( {
+} )
 export default class TermBox extends Vue {
+
+	get primaryLanguageName(): string {
+		return factory.getLanguageNameLookup().getLanguageName( this.primaryLanguage );
+	}
+
+	get primaryLanguage() {
+		// TODO: I tried doing a mapState within the @Component annotation,
```
wiese: https://github.com/ffxsam/vue-typescript-cookbook#im-using-vuex-mapstate-or-mapgetters-and-typescript-is-saying-the-mapped-stategetters-dont-exist-on-this ?

src/components/TermBox.vue
```diff
@@ -1,21 +1,40 @@
...
+
+@Component( {
+	computed: mapState( NS_USER, {
+		primaryLanguage: 'primaryLanguage',
+	} ),
+} )
+export default class TermBox extends ( Vue as VueConstructor<TermboxBindings> ) {
+	public primaryLanguageName = '';
+
+	public mounted() {
```
wiese: "This hook is not called during server-side rendering."
https://vuejs.org/v2/api/#mounted

src/components/TermBox.vue
```diff
@@ -1,21 +1,40 @@
...
+		primaryLanguage: 'primaryLanguage',
+	} ),
+} )
+export default class TermBox extends ( Vue as VueConstructor<TermboxBindings> ) {
+	public primaryLanguageName = '';
+
+	public created() {
+		factory.getLanguageRepository()
+			.getLanguageName( this.primaryLanguage, this.primaryLanguage )
+			.then( ( languageName ) => {
```
wiese: This will result in an update to `this.primaryLanguageName` _asynchronously_ which is not a problem in the client because resolution should be so quick that the visual effect can be disregarded, but creates a major problem in the server - as there is no way to wait for it this way.
I feel we may be better off with populating the store on dispatch and merely from it reading here.

Did some reading...
[0] https://ssr.vuejs.org/guide/data.html#data-store
[1] https://ssr.vuejs.org/guide/data.html#server-data-fetching
[2] https://ssr.vuejs.org/guide/routing.html#routing-with-vue-router
[3] https://github.com/iamfunsho/vue-ssr-example/blob/master/src/entry-server.js
[4] https://vuejsdevelopers.com/2017/12/11/vue-ssr-router/
[5] https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

PR #56, opened 2018-11-20
Configuration code style

wiese:
Let tslint check our JS (configuration) files in the project root, too.
With the release of palantir/tslint#3641 (I subscribed) the manual [prototypical] re-specification of `jsRules` will hopefully be obsolte soon.

Consistently indent JSON files. This goes against the tab indentation we
do for program files but allows for configuration files modified by
programs (e.g. npm using JSON.stringify() for package.json) to stay
consistent in the long run.

### DIFF
--- a/tsconfig.all.json
+++ b/tsconfig.all.json
@@ -5,15 +5,15 @@
     "module": "esnext",
     "sourceMap": true,
     "typeRoots": [
-		"./src/types",
-		"./src/types/client",
-		"./src/types/server",
-		"./node_modules/@types"
+      "./src/types",
+      "./src/types/client",
+      "./src/types/server",
+      "./node_modules/@types"
     ],
     "types": [
-		"webpack-env",
-		"jest",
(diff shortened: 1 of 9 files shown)


tsconfig.all.json
```diff
@@ -5,31 +5,33 @@
...
     ],
     "lib": [
       "esnext",
       "es5",
       "dom",
       "dom.iterable",
       "scripthost"
-    ]
+    ],
+    "allowJs": true
```
jakobw: nice

wiese:
https://github.com/palantir/tslint/pull/3641 was released in tslint 5.12.0 (comes with @vue/cli-plugin-typescript@3.2.0) - unfortunately [this comment](https://github.com/palantir/tslint/pull/3641#issuecomment-445478757) is right in that `extends` is not taken into account when building the jsRules so we can't switch it out just yet - I'll keep watching.

PR #57, opened 2018-11-20
Add expand collapse-all-lang image

jakobw:


wiese:
What is the rationale behind "naming by purpose" instead of a filename describing what we see in the picture? [Cf.](https://github.com/wmde/wikibase-termbox/blame/master/assets/README.md#L13) Should we re-think the rule?

wiese:
Replaces #41 I suppose.

PR #58, opened 2018-11-20
[WIP] server: add ability to load a language name (i18n)

wiese:
Bug: T209287


wiese:
Will be redone per `LanguageTranslationRepository` in #71

PR #67, opened 2018-11-21
[WIP] mv api request from TermboxHandler to store action

wiese:


tests/unit/store/entity/actions.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+			const entity = new FingerprintableEntity( entityId, {}, {}, {} );
+			factory.setEntityRepository( {
+				getFingerprintableEntity: ( thisEntityId: string ) => {
+					expect( thisEntityId ).toBe( entityId );
+					return Promise.resolve( entity );
+				},
+			} );
+			const context = newMinimalStore( {
+				commit: ( mutationName: string, thisEntity: FingerprintableEntity ) => {
```
jakobw: I know that we already have similar tests with the assertions in the mock callback, but I don't think they read very nicely. What do you think about
```typescript
			const commitMock = jest.fn();
			const context = newMinimalStore( {
				commit: commitMock,
			} );
			const action = actions[ ENTITY_INIT ] as any; // TODO

			action( context, entityId ).then( () => {
				expect( commitMock ).toBeCalledWith(
					ENTITY_INIT_MUTATION,
					entity,
				);
				done();
			} )
```

<sup>(I would've done this as an inline suggestion but I don't know how to make it multi line)</sup>

wiese: Using this for `language/actions.spec.ts`

PR #71, opened 2018-11-22
create language store, init it from App via actions

wiese:


wiese:
"works". Two broken suites: buildApp.spec.ts & init.spec.ts. Suggestion: get this all the way up to show-user-language (#49), final touches up there.

PR #73, opened 2018-11-22
server: add ability to load language names

wiese:
Bug: T209287

### DIFF
--- a/src/server-entry.ts
+++ b/src/server-entry.ts
@@ -7,10 +7,16 @@ import mwbot from 'mwbot';
 import EntityInitializer from './common/EntityInitializer';
 import getChildComponents from './common/getChildComponents';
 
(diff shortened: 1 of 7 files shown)


src/server-entry.ts
```diff
@@ -7,10 +8,16 @@ import mwbot from 'mwbot';
 import EntityInitializer from './common/EntityInitializer';
 import getChildComponents from './common/getChildComponents';
 
-factory.setLanguageTranslationRepository( new WikibaseApiLanguageRepository() );
+const WIKIBASE_REPO_API = config.getWikibaseRepoApi(); // TODO test
```
wiese: I don't particularly like this, and - as it stands - does not even help us to mock the value in tests (now dynamically setting the value from config there) but I see this as a separate challenge.

PR #74, opened 2018-11-23
Fix buildApp.spec

jakobw:


wiese:
s/does not even take a param any more/does not even _need_ a param any more/

jakobw:
@wiese thanks for the comment. It actually didn't make a lot of sense to test obsolete code. I took the chance to extract some untested code that the two entry scripts shared into it.

tests/unit/common/buildApp.spec.ts
```diff
@@ -1,25 +1,47 @@
...
+
+const mockGetChildComponents = jest.fn();
+jest.mock( '@/common/getChildComponents', () => ( {
+	__esModule: true,
+	default: ( app: any ) => mockGetChildComponents( app ),
+} ) );
+
+jest.mock( '@/store', () => ( {
+	__esModule: true,
+	default: {},
```
wiese: Witchcraft, took me a while. The value of this here should actually be the same object also passed to `mockAsyncData1` & `mockAsyncData2`'s first parameter, right?

jakobw: Yes, I'll create a variable for it so it's more obvious.

tests/unit/common/buildApp.spec.ts
```diff
@@ -1,25 +1,47 @@
...
-import FingerprintableEntity from '@/datamodel/FingerprintableEntity';
+import App from '@/components/App.vue';
+
+const mockGetChildComponents = jest.fn();
+jest.mock( '@/common/getChildComponents', () => ( {
+	__esModule: true,
+	default: ( app: any ) => mockGetChildComponents( app ),
+} ) );
+
+jest.mock( '@/store', () => ( {
```
wiese: Assuming that `jest.resetAllMocks()` or the likes of which are called on a per-file basis.

PR #81, opened 2018-11-27
nodemon: update to lose malicious package

wiese:
nodemon is a tool that helps develop node.js based applications by
automatically restarting the node application when files change.
This updates it to the latest version to fix a security problem
https://github.com/remy/nodemon/issues/1451

Dependency tree before was:
```
$ docker-compose run --rm node npm ls flatmap-stream
wikibase-termbox@0.1.0 /app
`-- nodemon@1.18.4
  `-- pstree.remy@1.1.0
    `-- ps-tree@1.1.0
      `-- event-stream@3.3.6
        `-- flatmap-stream@0.1.1      <- https://www.npmjs.com/advisories/737


wiese:
If you have time to spare, take a look at the [original issue](https://github.com/dominictarr/event-stream/issues/116) for some open source philosophy.

PR #86, opened 2018-11-28
InvalidRawEntityException: rm unused file

wiese:
Remove unused file containing an exception class definition not matching
its file name.


bitPogo:
check

PR #90, opened 2018-12-03
Termbox: Paint it

wiese:
Initial styling for termbox when used in `&useformat=mobile`.
This omits layout details, responsiveness to create the playing field for more fine-grained changes. This also omits the development version (`public/index.html`) which could do with _some_ environment for the component to be rendered in (inherit wikibase styles / just use a box / ...) - let's discuss before we invest in either direction.

Please see the individual commits for modular changes and reasoning.

Pin in wikibase needs to be updated after main line merge for changes to come into effect in SSR and CSR ([cf.](https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/476541)).

Bug: T207469


src/components/App.vue
```diff
@@ -35,4 +35,7 @@ export default class App extends Vue {
 </script>
 
 <style lang="scss">
+.wikibase-entitytermsview {
+	@import 'reset-css';
```
bitPogo: Looks like black magic

src/filter/message.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * TODO dynamic messages
+ * When state (user language) comes into play, using a computed will likely be
+ * the more correct™ approach - https://stackoverflow.com/a/42829248
+ */
+export default ( key: string ): string => {
+	if ( key === 'wikibase-termbox-alias-separator' ) {
```
bitPogo: Maybe it's good to document this variance

src/components/TermBox.vue
```diff
@@ -1,19 +1,20 @@
 <template>
-	<div class="wikibase-termbox-primaryLanguage">
+	<!-- TODO dynamize wikibase-termbox--primaryLanguage for T205261 -->
+	<div class="wikibase-termbox wikibase-termbox--primaryLanguage">
```
bitPogo: 'wikibase-termbox' - depends where the other components are located should be probably should be move one level up.

src/components/App.vue
wiese: https://meyerweb.com/eric/tools/css/reset/

src/components/TermBox.vue
wiese: I feel the components should be self-contained and found the **B**EM block class present only with a modifier attached problematic.
Compare "Under the hood" http://getbem.com/introduction/

PR #100, opened 2018-12-05
Get language names by wb.WikibaseContentLanguage

bitPogo:


bitPogo:
Note: There is also a patch for wikibase needed, which enables uls functionality also for mobile.

tests/unit/client/data-access/UlsLanguageRepository.spec.ts
```diff
@@ -1,26 +1,36 @@
 import UlsLanguageTranslationRepository from '@/client/data-access/UlsLanguageTranslationRepository';
+import { StringTMap } from '@/datamodel/LanguageTranslations';
+import MwWindow from '@/client/MwWindow';
 
-function newUlsLanguageRepository( getLanguageNameByCode: ( langCode: string ) => string, langChain: string[] ) {
-	return new UlsLanguageTranslationRepository( getLanguageNameByCode, langChain );
+class MockWbWikibaseContentLanguages {
```
wiese: This in `MwWindow.ts` would give us the type support for which we went with TS.

src/client/data-access/UlsLanguageTranslationRepository.ts
```diff
@@ -1,35 +1,18 @@
...
 	}
 
-	private getLanguagesNamesInLanguage(): StringTMap<string> {
-		const languageNamesInUserLang = {} as StringTMap<string>;
-		this.languagesAtEntity.forEach( ( item: string ) => {
-			languageNamesInUserLang[ item ] = this.getLanguageNameInUserLang( item );
-		} );
-		return languageNamesInUserLang;
+	private getLanguagesNames(): StringTMap<string> {
+		return new ( ( window as MwWindow ).wb.WikibaseContentLanguages.prototype.constructor )().getAllPairs();
```
wiese: The content languages are a service I see us reusing to implement the LANGUAGE_INIT action so I don't believe it would be premature to inject it.

wiese: It is still missing in dev mode (`src/mockup-entry.ts`).

tests/unit/client/data-access/UlsLanguageRepository.spec.ts
bitPogo: I agree on this totally, but ts make nearly impossible to access the constructor if we define it as a type at wb. The solutions for that are very hacky and it's just a way to work around ts and not work on the 'real' problem.

src/client/data-access/UlsLanguageTranslationRepository.ts
bitPogo: done

bitPogo:
Probably it's turned red because the travis problem?

src/client/data-access/UlsLanguageTranslationRepository.ts
bitPogo: If we inject this in the entry point, we should do the same for https://github.com/wmde/wikibase-termbox/blob/master/src/client/data-access/EntityRepository.ts (maybe also https://github.com/wmde/wikibase-termbox/blob/master/src/client/init.ts),  because I like the idea that in the end is one clear way where MwWindow comes from, otherwise we spread this in n+1 ways.

tests/unit/client/data-access/UlsLanguageTranslationRepository.spec.ts
```diff
@@ -0,0 +1,38 @@
+import UlsLanguageTranslationRepository from '@/client/data-access/UlsLanguageTranslationRepository';
+import { StringTMap } from '@/datamodel/LanguageTranslations';
+import MwWindow from '@/client/MwWindow';
+
+class MockWbWikibaseContentLanguages {
+	public static translations: StringTMap<string>;
+	constructor() {
```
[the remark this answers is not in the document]
bitPogo: done

src/mockup-entry.ts
```diff
@@ -8,8 +10,16 @@ import MwWindow from './client/MwWindow';
 	hook: () => new ImmediatelyInvokingEntityLoadedHookHandler( entity ),
 };
 
+class MockupWikibaseContentLanguages {
+	constructor() {
```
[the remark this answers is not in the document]
bitPogo: done

src/client/data-access/UlsLanguageTranslationRepository.ts
bitPogo: Done

tests/unit/client/data-access/UlsLanguageTranslationRepository.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+function mockMwEnv( translations: any ) {
+	( window as MwWindow ).wb = {
+		WikibaseContentLanguages: translations,
+	};
+}
+
+describe( 'UlsLanguageTranslationRepository', () => {
+
+	it( 'gets the language name from the given function', () => {
+		const langTranslation = { en: 'English' };
```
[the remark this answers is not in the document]
bitPogo: done

src/client/MwWindow.ts
```diff
@@ -11,8 +11,12 @@ interface MediaWiki {
 	config: MwConfig;
 }
 
+export interface WikibaseContentLanguages {
+	getAllPairs: () => any;
+}
+
 interface Wikibase {
-	getLanguageNameByCode( languageCode: string ): string;
+	WikibaseContentLanguages: any;
```
jakobw: I just learned that you can define construct signatures to hint against classes in TS! https://www.typescriptlang.org/docs/handbook/interfaces.html#difference-between-the-static-and-instance-sides-of-classes

This could then become something like
```suggestion
	WikibaseContentLanguages: {
		new(): WikibaseContentLanguages;
	};
```

tests/unit/client/data-access/UlsLanguageTranslationRepository.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+		this.translations = translations;
+	}
+
+	public getAllPairs(): StringTMap<string> {
+		return this.translations;
+	}
+}
+
+describe( 'UlsLanguageTranslationRepository', () => {
+	const langTranslation = { en: 'English' };
```
jakobw: These two variables `langTranslation` and `translator` should move into the `it`.

tests/unit/client/data-access/UlsLanguageTranslationRepository.spec.ts
```diff
@@ -0,0 +1,28 @@
+import UlsLanguageTranslationRepository from '@/client/data-access/UlsLanguageTranslationRepository';
```
jakobw: This whole test got so much better, now that its dependency is injected! <3

tests/unit/client/data-access/UlsLanguageTranslationRepository.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+import { StringTMap } from '@/datamodel/LanguageTranslations';
+import { WikibaseContentLanguages } from '@/client/MwWindow';
+
+class MockWbWikibaseContentLanguages {
```
jakobw: The class isn't necessary. If we inject a plain object literal, the whole test becomes super simple:
```TypeScript
import UlsLanguageTranslationRepository from '@/client/data-access/UlsLanguageTranslationRepository';

describe( 'UlsLanguageTranslationRepository', () => {
	it( 'gets the language names from the given language translation service', () => {
		const langTranslation = { en: 'English' };
		const translator = {
			getAllPairs() {
				return langTranslation;
			},
		};

		return (
			new UlsLanguageTranslationRepository( translator )
		).getLanguagesInLanguage( 'en' ).then( ( name ) => {
			expect( name ).toEqual( {
				en: langTranslation,
			} );
		} );
	} );
} );
```

bitPogo: done

wiese:
He @bitPogo,
care to rebase this on top of #105 and merge this together? [See](https://github.com/wmde/wikibase-termbox/pull/105/files#diff-e0af28ba1e3f656bdd261302b1c324cfR15)

PR #105, opened 2018-12-07
languages: interface, server implementation

wiese:
Provide languages and their directionality to the application.

Use this as base to set languages in the client.

Bug: T210407


wiese:
Rebased

src/server-entry.ts
```diff
@@ -1,34 +1,45 @@
...
-		apiBot,
-	),
-);
-factory.setEntityRepository(
-	new MwBotWikibaseRepo(
-		apiBot,
-		new EntityInitializer(),
-	),
-);
+	const languageRepo = new WikibaseContentLanguagesRepo(
```
wiese: Had we continued doing this outside the export, all requests would have a shared the same language API response. This could be, carefully crafted, a performance optimization in the future but should be an active decision. (Not currently possible for this scenario with the API response containing request-specific content [translation])

wiese:
Rebased

src/server/data-access/WikibaseContentLanguagesRepo.ts
```diff
@@ -0,0 +1,66 @@
...
+			} );
+
+			return languages;
+		} );
+	}
+
+	public getLanguagesInLanguage( inLanguage: string ): Promise<LanguageTranslations> {
+		this.loadWbContentLanguages( inLanguage );
+		return this.wbContentLanguages.promise.then( ( contentLanguages: WikibaseApiContentLanguages ) => {
+			const translations = {} as any; // TODO why is multi [inLanguage][languageCode] not valid? not ltr?
```
jakobw: Not sure if I understand it correctly but the following looks cleaner IMO

wiese: The todo is a copy from the previous incarnation of this class. If there is a cleaner way of writing this I'm happy to see it go.

wiese:
Updated to address @jakobw's valid concerns.

wiese:
Rebased

PR #106, opened 2018-12-07
Render termboxes in all languages

wiese:
Render a termbox in user language and then individual termboxes for all
remaining languages.
By no means complete but acts as a skeleton for smaller construction
sites that now can be individually addresses in detail.

Bug: T205261


tests/edge-to-edge/server/app.spec.ts
```diff
@@ -166,21 +172,32 @@ describe( 'Termbox SSR', () => {
 
 			const $dom = getDomFromMarkup( response.text );
 
-			expect( $dom.querySelectorAll( '.wikibase-termbox' ).length ).toBe( 1 );
+			const termboxVersions = $dom.querySelectorAll( '.wikibase-termbox' );
+
+			// 3 hard coded languages currently in src/store/language/actions.ts
+			expect( termboxVersions.length ).toBe( 3 );
```
bitPogo: Maybe Object.keys( response.wbcontentlanguages ).length instead?

src/components/TermBox.vue
```diff
@@ -83,20 +88,24 @@ export default class TermBox extends ( Vue as VueConstructor<TermboxBindings> )
 	}
 
 	get hasAliases(): boolean {
-		return !( this.entityAliases( this.primaryLanguage ) == null );
+		return !( this.entityAliases( this.language.code ) == null );
```
jakobw: Rule 3 applies, but this method looks odd. Shouldn't it be
```suggestion
		return this.entityAliases( this.language.code ) !== null;
```

src/components/TermBox.vue
```diff
@@ -1,9 +1,11 @@
 <template>
-	<!-- TODO dynamize wikibase-termbox--primaryLanguage for T205261 -->
-	<div class="wikibase-termbox wikibase-termbox--primaryLanguage">
+	<div class="wikibase-termbox"
+		:class="{ 'wikibase-termbox--primaryLanguage': isInUserLanguage }">
```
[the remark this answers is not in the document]
wiese: @jakobw explained to me what he means: wrapping the TermBox that is in primaryLanguage in a dedicated element inside App and not burdening TermBox with that task at all.
What I do not like about that approach, from the purely practical side of things, is the question where the CSS would then be specified. TermBox would specify CSS for a selector chain that it itself only controls partially, making the component non-portable.
From the conceptional side of things I'm also skeptical - to me it feels like a performance optimization ("we already know that's the primaryLanguage in App"), ignoring the fact that it is the TermBox component that should control its characteristics, and, of course, needs access to the data (store) to decide on it. 
What I could imagine as a compromise to add another prop to TermBox, making TermBox less of a self-reliant component and the primaryLanguage (name to be discussed) configurable from the outside.

wiese:
Closing this as it makes little sense until #105 is in.

tests/edge-to-edge/server/app.spec.ts
wiese: I intentionally hard-coded this to break as soon as the list is dynamized (e.g. https://phabricator.wikimedia.org/T209288).

PR #107, opened 2018-12-10
production build: ensure 404s are reported correctly

wiese:
In production builds, with minification enabled, exceptions that were
created inside the server bundle could be neither analyzed by class (as
classes loaded inside and outside the bundle are effectively seen as
two independent ones) nor class name (as minification caused the name to
be shortened).
This introduces a dedicated error response object for this rare scenario
for the server to be able to further analyze the reason of errors from
within the bundle, e.g. to tell apart a server error from an entity that
can not be found.
This also disables minification in the node environment of which there is
no clear advantage (size is not a factor) but which deprives us of the
ability to identify the error response by name.

Bug: T211573


vue.config.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
 		},
 		optimization: {
 			splitChunks: undefined,
+			minimize: !TARGET_NODE, // needed for comparison of `.constructor.name`s
```
wiese: Maybe [namedModules](https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a#166e) or namedChunks would suffice?

PR #113, opened 2018-12-12
tests: hide production tip in e2e tests

wiese:
edge-to-edge tests are simulating actual requests against the server
and are - by definition - run in a development context. While the
production tip is valuable for the server (should we bork config) it
makes no sense to show it during tests.

### DIFF
--- a/tests/edge-to-edge/server/app.spec.ts
+++ b/tests/edge-to-edge/server/app.spec.ts
@@ -5,6 +5,15 @@ import nock from 'nock';
 import { config } from '@/server/TermboxConfig';
 import 'jest-dom/extend-expect';
 import HttpStatus from 'http-status-codes';
+import Vue from 'vue';
+
+/**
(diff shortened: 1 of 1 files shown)


tests/edge-to-edge/server/app.spec.ts
```diff
@@ -5,6 +5,15 @@ import nock from 'nock';
 import { config } from '@/server/TermboxConfig';
 import 'jest-dom/extend-expect';
 import HttpStatus from 'http-status-codes';
+import Vue from 'vue';
```
jakobw: Wouldn't `tests/config/setup.ts` be a better place for this?

wiese: In contrast to real HTTP request hitting express (inside of which env is pretty much empty, i.e. not in `NODE_ENV: 'development'` in particular), the server-side code of the edge-to-edge tests is executed with the local env, i.e. development mode, which triggers the warnings inside vue(x).

PR #115, opened 2018-12-13
env: explicitly set variables relevant for server process

wiese:
This resolves a test todo (mocking based on .env) and allows us to
specify variables relevant to the runtime on server start instead of on
server/[blubber] image build.
This also introduces the concept of services that are reusable accross
multiple requests (e.g. mwbot) and provides a way of passing it to the
individual process handler.

Bug: T211675


src/server-entry.ts
```diff
@@ -1,32 +1,32 @@
...
 import EntityInitializer from './common/EntityInitializer';
 import BundleBoundaryPassingException, { ErrorReason } from '@/server/exceptions/BundleBoundaryPassingException';
-import TermboxRequest from '@/common/TermboxRequest';
 import EntityNotFound from '@/common/data-access/error/EntityNotFound';
+import SsrRequest from './server/SsrRequest';
 
-const apiBot = new mwbot( {
-	apiUrl: config.getWikibaseRepoApi(), // TODO test
-} );
+export default ( ssrRequest: SsrRequest ) => {
```
jakobw: I'm not strongly opposing, but it does not sound quite right to make the config part of the request. It also means that any service dependent on its values can only be created within this function, which I find icky considering earlier today we said that `mwbot` looks like a good candidate to be shared between requests. Only having the API URL available here wouldn't make it impossible, but likely more complicated.

manicki:
Jenkins/Blubber approve: https://gerrit.wikimedia.org/r/#/c/wikibase/termbox/+/479536/

src/server-entry.ts
wiese: While I agree with you that having the config appearing to be part of the request (maybe poor naming contributed to this), it is not quite what's happening per se.
As I understand it, please correct me, there is only one way to get information (_context_) into the bundle - through the first parameter of `serverBundle`'s `renderToString` method (that is src/server-entry.ts).
Consequently, if we _want_ to pass information into the bundle it has to go through here. Do you see an alternative?

wiese:
Updated

tests/unit/server-entry.spec.ts
```diff
@@ -2,14 +2,56 @@ import serverEntry from '@/server-entry';
...
 const mockBuildApp = jest.fn();
 jest.mock( '@/common/buildApp', () => ( {
 	__esModule: true,
 	default: ( termboxRequest: TermboxRequest ) => mockBuildApp( termboxRequest ),
 } ) );
 
+const getMediawikiBot = jest.fn();
+function getMockBundleRendererServices() {
+	const services = {} as BundleRendererServices;
+	Object.defineProperty( services, 'mediawikiBot', {
```
jakobw: I tried and failed to get rid of this while keeping the spy on the property access :(
I found out that it is possible to mock/spy on getters which is nice. Doesn't help in this case, though.

Maybe an alternative would have been to assert that the same mwbot instance was passed to the constructors of LanguagesRepo and EntityRepo instead. OTOH I don't think it's worth the hassle to change this now.

.pipeline/blubber.yaml
```diff
@@ -20,3 +21,25 @@ variants:
...
+        source: /opt/lib/serverBuild
+        destination: /srv/service/serverBuild
+      - from: build-server
+        source: /opt/lib/serverDist
+        destination: /srv/service/serverDist
+    lives:
+      in: /srv/service
+    node: { env: production }
+    # we have to run npm in development mode to get all the dependencies we need to build-server
+    # node: { env: production }
```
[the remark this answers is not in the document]
wiese: This and the previous line are actually resolved through the addition of tslib (#122).

README.md
```diff
@@ -49,9 +49,28 @@ Instructions above will gradually be migrated to use blubber.
...
+blubber .pipeline/blubber.yaml production > Dockerfile
+docker build -t wmde/wikibase-termbox-production .
+docker run --rm -p "3030:3030" -e WIKIBASE_REPO_API=http://default.web.mw.localhost/mediawiki/api.php -e SSR_PORT=3030 --network addshoremediawikidockerdev_default --network-alias node-ssr --link nginx-proxy:default.web.mw.localhost wmde/wikibase-termbox-production
+```
+
+or, more conveniently, `docker-compose run -T ssr node serverBuild/server/server.js`
+
+### Running for production
+
+> THIS IS WIP - do not use!
```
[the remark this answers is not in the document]
wiese: This is to address your "devel" concern, too. This patch is definitely not bringing the blubber configuration to a conclusion but enables the application to be used inside of it - the blubber touches are just to showcase it actually functions as expected (see PR description). I'll remove it from this patch however as it does not seem enough to know that it's unused.

manicki:
Very cool!
I've synced the gerrit "mirror" of this repo, and Jenkins is still positive: https://gerrit.wikimedia.org/r/#/c/wikibase/termbox/+/481131/ !

PR #116, opened 2018-12-14
Inject hook handler dependency into client/EntityRepository

jakobw:


src/client-entry.ts
```diff
@@ -5,14 +5,16 @@ import TermboxRequest from '@/common/TermboxRequest';
 import { factory } from '@/common/TermboxFactory';
 import UlsLanguageTranslationRepository from '@/client/data-access/UlsLanguageTranslationRepository';
 import EntityRepository from '@/client/data-access/EntityRepository';
-import MwWindow from './client/MwWindow';
+import MwWindow from './client/mediawiki/MwWindow';
```
bitPogo: As Pablo mentioned earlier in one of my commits - can we use relative or absolute paths and not mix it? Or at least set up a rule, when we are allowed to use relative paths?

jakobw: Good point, fixed!

src/client-entry.ts
```diff
@@ -5,14 +5,16 @@ import TermboxRequest from '@/common/TermboxRequest';
 import { factory } from '@/common/TermboxFactory';
 import UlsLanguageTranslationRepository from '@/client/data-access/UlsLanguageTranslationRepository';
 import EntityRepository from '@/client/data-access/EntityRepository';
-import MwWindow from './client/MwWindow';
+import MwWindow from '@/client/mediawiki/MwWindow';
```
wiese: I vaguely remember the discussion when we introduced this. 
I'm still fine with inline casting when using e.g. window in a way that we "hope it to be" a MwWindow.
But this move into a dedicated folder makes me question why this is not in a (obviously client-only) `.d.ts` folder instead...

src/client-entry.ts
```diff
@@ -5,14 +5,16 @@ import TermboxRequest from '@/common/TermboxRequest';
...
 
 Vue.config.productionTip = false;
 
 factory.setLanguageTranslationRepository(
 	new UlsLanguageTranslationRepository( ( window as MwWindow ).wb.getLanguageNameByCode ),
 );
-factory.setEntityRepository( new EntityRepository() );
+factory.setEntityRepository( new EntityRepository(
+	( window as MwWindow ).mw.hook( 'wikibase.entityPage.entityLoaded' ),
```
wiese: Not that it makes any difference (+ rule n°3) but in an ideal world we would probably have a list of named constants of available hooks.

jakobw: Good idea. I added `src/client/mediawiki/Hooks.ts`

PR #117, opened 2018-12-14
Disallow inferrable type declarations

jakobw:
Unfortunately, this only works for variables and parameters (but not return types?), and only for number, string, or boolean. https://palantir.github.io/tslint/rules/no-inferrable-types/

I stumbled across this looking for something that can help us prevent unnecessary casting, which I saw in quite some places, but as far as I can tell there is no such thing.

### DIFF
--- a/src/store/entity/index.ts
+++ b/src/store/entity/index.ts
@@ -11,7 +11,7 @@ const state: Entity = {
 	aliases: {},
 };
 
-const namespaced: boolean = true;
+const namespaced = true;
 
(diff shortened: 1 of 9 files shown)


jakobw:
From the rule description:
> Explicit types where they can be easily inferred by the compiler make code more verbose.

wiese:
I read the thought but to me it does not sound ("easily") like a gain, to the contrary actually - I'm fine to try it, however.

PR #120, opened 2018-12-17
Make assets avaible for dev and wb

bitPogo:


bitPogo:
The branch example-asset-import hold the sample implementation for usage.
**Note:**  if you make a build the svg's will be included and added in a separate img directory in dist

bitPogo:
changed

package-lock.json
```diff
@@ -2115,7 +2115,7 @@
     },
     "array-equal": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
```
wiese: That has nothing to do with the story and looks a bit dangerous - please remove.

.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules
 dist/*
 !dist/wikibase.termbox.main.js
 !dist/wikibase.termbox.main.css
+!dist/img/*
```
wiese: We would want this (if it were a thing) to be ignored (L3 does that already) - in wikibase we can use the `assets` folder.

package-lock.json
```diff
@@ -3170,7 +3170,7 @@
     },
     "callsites": {
       "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
```
wiese: That has nothing to do with the story and looks a bit dangerous - please remove.

.gitignore
bitPogo: img we be there if you make the build with the example branch. In this pull request there is nothing which links to this asset, therefore it will not compiled.
I agree it is not the best way to have assets at least twice, but so far I have no solution to provide a better way then this.

package-lock.json
bitPogo: done

vue.config.js
```diff
@@ -40,6 +41,16 @@ module.exports = {
...
 				] );
+
+			config.module
+				.rule('images')
+				.test( /\.(png|jpe?g|gif|svg)(\?.*)?$/ )
+				.use( 'url-loader' )
+				.loader( 'url-loader' )
+				.options( {
+					limit: -1,
+					name: assetsPath + '/[name].[ext]',
```
jakobw: If we do not include `[path]` here somehow, this will only work if all assets are directly in `/assets`, without sub-directories. Was that an active decision?

package.json
```diff
@@ -48,6 +48,7 @@
     "node-sass": "^4.9.0",
     "nodemon": "^1.18.7",
     "reset-css": "^4.0.1",
+    "sass-inline-svg-utf8": "^1.0.5",
```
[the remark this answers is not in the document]
jakobw: Uh, cool. This comment was not meant to make it here. Apparently I had started reviewing an old version of this patch and never finished it. Please ignore.

vue.config.js
wiese: Was a "good enough" decision (no sub folders ATM) but would definitely benefit from future-proofing this.

PR #121, opened 2018-12-18
app.ts: hint type for server app

wiese:
To ease use, more self-explanatory code.

### DIFF
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -1,5 +1,5 @@
 import fs from 'fs';
-import express from 'express';
+import express, { Request, Response } from 'express';
 import { createBundleRenderer } from 'vue-server-renderer';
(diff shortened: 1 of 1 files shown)


src/server/app.ts
```diff
@@ -14,7 +14,7 @@ const renderer = createBundleRenderer(
 	{ runInNewContext: false },
 );
 
-app.get( '/termbox', ( request, response ) => {
+app.get( '/termbox', ( request: Request, response: Response ) => {
```
jakobw: Hm, my IDE seems to guess the correct type, but then still gives different hints once the type is made explicit. 

req

wiese: I won a XMLHttpRequest which was less than helpful.

PR #122, opened 2018-12-18
tslib: ensure presence in prod installation

wiese:
We run typescript with the configuration `importHelpers: true`,
consequently the compiler will inject helper functions into the
generated output that are called at run-time. These helper functions
reside in the tslib library which consequently has to be installed.
So far it has only been installed as a 3rd party dependency of some of
our dev dependencies. Consequently running the server with only prod
dependencies installed, as we plan for production, failed.
See
https://blog.mariusschulz.com/2016/12/16/typescript-2-1-external-helpers-library
https://www.npmjs.com/package/tslib

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15148,8 +15148,7 @@
     "tslib": {
       "version": "1.9.3",
       "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.9.3.tgz",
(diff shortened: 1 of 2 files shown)


wiese:
Showcase https://drive.google.com/file/d/1AHY5SDg3ZnTL5AZGyTCQmdKvry17F9z0/view

PR #123, opened 2018-12-18
Remove unnecessary type casts

jakobw:
Explicit type hints are fine and can be helpful even if they are technically redundant, but unnecessary casts are confusing, because they give the impression that we're forcing an object into a type that it isn't.


wiese:
https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html#assertion-considered-harmful

PR #125, opened 2018-12-18
Add edit link url

jakobw:


src/store/RootState.ts
```diff
@@ -0,0 +1,3 @@
+export default interface RootState {
```
jakobw: Will move this into a links module, as discussed.

tests/edge-to-edge/server/app.spec.ts
```diff
@@ -202,11 +207,12 @@ describe( 'Termbox SSR', () => {
...
+		const editLink = '/some';
 
 		nockSuccessfulLanguageLoading( language );
 		nockSuccessfulEntityLoading( entityId );
 
-		request( app ).get( '/termbox' ).query( { entity: entityId, language } ).then( ( response ) => {
+		request( app ).get( '/termbox' ).query( { entity: entityId, language, editLink } ).then( ( response ) => {
 			expect( response.status ).toBe( HttpStatus.OK );
 
 			const $dom = getDomFromMarkup( response.text );
```
wiese: Should we not assert that the link appears in the mark-up?

src/components/TermBox.vue
```diff
@@ -14,7 +14,11 @@
 				<p class="wikibase-termbox__aliases wikibase-termbox__aliases--placeholder" v-else>?</p>
 			</div>
 		</div>
-		<div class="wikibase-termbox__action"></div>
+		<div class="wikibase-termbox__actions">
+			<div class="wikibase-termbox__action--edit">
+				<a :href="editLinkUrl">PEN</a>
```
wiese: This is not tested for in the component unit test.

tests/edge-to-edge/server/app.spec.ts
jakobw: yes! done

jakobw:
> How do we sync this with the pen CSS?

This adds an invisible link so it could be merged as is?

PR #130, opened 2018-12-19
Messagekey json generator

bitPogo:


jakobw:
Thanks for sharing! I like the idea of automating this step, although I don't really see the need for it at the moment, as long as we only have very few messages. Closing this to lower our open PR count. Feel free to open this again, if necessary!

PR #135, opened 2018-12-20
Use conventional test syntax instead of magic "resolves"

jakobw:
Since there was another assertion in the test, and it did not actively
wait for the promise to be resolved, the test passed without evaluating
the second assertion.


wiese:
Went into #133

PR #140, opened 2018-12-21
Flip pen for rtl languages

jakobw:
https://phabricator.wikimedia.org/T211589

pen-left-pen-right


src/components/EditPen.vue
```diff
@@ -25,6 +25,10 @@ export default class EditPen extends Vue {
 		width: 48px;
 		height: 48px;
 		display: block;
+
+		.rtl & {
+			transform: scaleX(-1);
```
wiese: https://caniuse.com/#feat=transforms2d

jakobw:
Updated!
* now using `dir` instead of a class + the respective attribute selector in `EditPen.vue`
* using the `getLanguageByCode` getter
* hopefully made the `store` access in tests a little more obvious

wiese:
Flip based on `dir` seems to work in IE10, too.
virtualbox_ie10 - win7_02_01_2019_12_19_41

PR #143, opened 2018-12-21
Show language used in dev mode in URL

wiese:
So that it's clear which language we look at and can easily change it.


src/mockup-entry.ts
```diff
@@ -5,11 +5,14 @@ import * as languageMap from '@/mock-data/data/en_lang_data.json';
 import * as directionalities from '@/mock-data/data/en_lang_dir_data.json';
 import ImmediatelyInvokingEntityLoadedHookHandler from '@/mock-data/ImmediatelyInvokingEntityLoadedHookHandler';
 import MwWindow from '@/client/mediawiki/MwWindow';
+import getOrEnforceUrlParameter from './mock-data/getOrEnforceUrlParameter';
+
+const language = getOrEnforceUrlParameter( 'language', 'de' );
```
wiese: The hard-wired "language" value could definitely come from a central place (see ` TermboxHandler`, `QueryValidator`) but let's wait for things to blow over in the SSR request department before investing more into this.

PR #145, opened 2019-01-02
Configure supported browserlist for transpiling

jakobw:
Bug: T209930


.browserslistrc
```diff
@@ -0,0 +1,10 @@
+last 2 Chrome versions
+IE >=11
+last 2 Firefox versions
+Safari >=5.1
+Opera >=15
+iOS >=6.1
+Android >=4.1
+
+IE 10 # only basic support according to the mw browser compatibility matrix.
+# TODO: how to separate CSS and JS browser support?
```
wiese: We could certainly pimp the `postcss.config.js` to merge `.browserslistrc` and the CSS-only browser list but natively I could not find the hybrid version [at first glance](https://github.com/browserslist/browserslist#user-content-queries).

wiese:
**Before (master)**
wikibase.termbox.main.css 4960 bytes
wikibase.termbox.main.js 112968 bytes

**After (browser-support)**
wikibase.termbox.main.css 5021 bytes
wikibase.termbox.main.js 111415 bytes

Did the JS get smaller?

jakobw:
I included `defaults`, and the build is still smaller than the original ಠ_ಠ

wiese:
Gives us the flex box support for IE10 we missing in #140 

virtualbox_ie10 - win7_03_01_2019_13_13_30

(I just now saw the margin problem in RTL, but this is not browser-specific. Are we relying on CSS Janus there?)

jakobw:
Yes, I believe CSS Janus is helping with the margins. At least on the screenshot in #140 it looked like everything is aligned correctly.

PR #150, opened 2019-01-07
Messages server

bitPogo:


src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,86 @@
...
+	public constructor( bot: mwbot, messageKeys?: string[] ) {
+		this.bot = bot;
+		if ( messageKeys === null || typeof messageKeys === 'undefined' ) {
+			this.messageKeys = this.getRequestedKeys();
+		} else {
+			this.messageKeys = messageKeys.join( '|' );
+		}
+	}
+
+	public getMessagesInLanguage( inLanguage: string|null ): Promise<MessageTranslationCollection> {
```
wiese: I did not try to figure out which version of which branch this change is following but the usefulness of invoking a method called `*InLanguage` without a language parameter escapes me. The parameter should be required.
If (speculating) another method, implementing that same interface, decides not to make use of that parameter, that is its (lonesome) decision.

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,86 @@
...
+import MessagesRepository from '@/common/data-access/MessagesRepository';
+import mwbot from 'mwbot';
+
+interface RespondMessage {
+	[ key: string]: string;
+}
+
+export default class MwBotWikibaseMessagesRepo implements MessagesRepository {
+	private bot: mwbot;
+	private messageKeys: string;
```
wiese:
```suggestion
	private messageKeys: array;
```
(explanation below)

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,86 @@
...
+		const messageCollection: MessageTranslationCollection = {};
+		messageCollection[ inLanguage ] = this.transformMessages( response );
+		return messageCollection;
+	}
+
+	private transformMessages( response: RespondMessage[] ): MessageCollection {
+		const messageList: MessageCollection = {};
+		response.forEach( ( value: { [key: string]: string } ) => {
+			if ( value.hasOwnProperty( 'missing' ) ) {
+				messageList[ value.name ] = value.name; // this is just the same behavior like mw
```
wiese: This situation can only occur if we have a message key configured that is not supported by mw. That is, this does not happen if a translation in a given language is missing in mw (in which case the English message is shown), but if the **key** is not present, right? So should this happen we are actually in a bad situation. I'm leaning towards actually throwing an exception.

src/common/MessageKeys.ts
```diff
@@ -0,0 +1,8 @@
+export enum MessageKeys {
+	EDIT = 'wikibase-edit',
+	AUGUST = 'august',
```
wiese: These were probably added for trying it out, but obviously have to be removed.

tests/edge-to-edge/server/app.spec.ts
```diff
@@ -209,6 +242,7 @@ describe( 'Termbox SSR', () => {
 		const editLink = '/some/' + entityId;
 
 		nockSuccessfulLanguageLoading( language );
+		nockSuccessfulMessagesLoading( 'de' );
```
wiese:
```suggestion
		nockSuccessfulMessagesLoading( language );
```

tests/unit/store/user/actions.spec.ts
```diff
@@ -28,6 +32,11 @@ describe( 'user/actions', () => {
 					inLanguage,
 					{ root: true },
 				);
+				expect( dispatchMock ).toBeCalledWith(
```
wiese: This should be obsoleted by some rebasing, I guess.

tests/edge-to-edge/server/app.spec.ts
```diff
@@ -21,6 +21,21 @@ const app = createApp( WIKIBASE_TEST_API_HOST + WIKIBASE_TEST_API_PATH );
...
+	"name": "wikibase-alias-edit-placeholder",
+	"normalizedname": "wikibase-alias-edit-placeholder",
+	"*": "Alias eingeben"
+}, {
+	"name": "wikibase-edit",
+	"normalizedname": "wikibase-edit",
+	"*": "bearbeiten"
+} ];
+
+const messageKeys = [
```
wiese: This needs to be in line with what is actually requested (edge-to-edge test).

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,86 @@
...
+export default class MwBotWikibaseMessagesRepo implements MessagesRepository {
+	private bot: mwbot;
+	private messageKeys: string;
+
+	public constructor( bot: mwbot, messageKeys?: string[] ) {
+		this.bot = bot;
+		if ( messageKeys === null || typeof messageKeys === 'undefined' ) {
+			this.messageKeys = this.getRequestedKeys();
+		} else {
+			this.messageKeys = messageKeys.join( '|' );
```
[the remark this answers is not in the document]
bitPogo: He does not (unfortunately). If the bot gets a array in this tries case, it fetch all messages, which is also a way to success, but I guess not the way we want.

src/server/data-access/MwBotWikibaseMessagesRepo.ts
bitPogo: right...and we can trust the mw fallback chain...

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,75 @@
+import TechnicalProblem from '@/common/data-access/error/TechnicalProblem';
+import MessageNotFound from '@/common/data-access/error/MessageNotFound';
+import MessageTranslationCollection from '@/datamodel/MessageTranslationCollection';
+import MessageCollection from '@/datamodel/MessageCollection';
+import MessagesRepository from '@/common/data-access/MessagesRepository';
+import mwbot from 'mwbot';
+
+interface RespondMessage {
+	[ key: string]: string;
```
jakobw: It seems that the individual messages returned from `query=allmessages` always have a `name`, `normalizedname`, `*`, and possibly `missing` property, so why not include that in the interface? :)

Also, I tripped over the name for a moment. This is an object that's part of the `allmessages` response, so how about something like `AllMessagesResponseMessage`? <sup>yes, not a great name either</sup>

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,75 @@
...
+		response: RespondMessage[],
+	): MessageTranslationCollection {
+		const messageCollection: MessageTranslationCollection = {};
+		messageCollection[ inLanguage ] = this.transformMessages( response );
+		return messageCollection;
+	}
+
+	private transformMessages( response: RespondMessage[] ): MessageCollection {
+		const messageList: MessageCollection = {};
+		response.forEach( ( value: { [key: string]: string } ) => {
```
jakobw: `value` here is of type `RespondMessage`, so the hint and the name could be better

bitPogo: done

src/server/data-access/MwBotWikibaseMessagesRepo.ts
bitPogo: done

tests/unit/server/data-access/MwBotWikibaseMessagesRepo.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+
+	describe( 'getMessagesInLanguage', () => {
+		it( 'creates a well-formed allmessages query with amlang', ( done ) => {
+			const inLanguage = 'de';
+			const bot = {
+				request: ( params: object ) => {
+					expect( params ).toEqual( {
+						action: 'query',
+						meta: 'allmessages',
+						ammessages: Object.values( MessageKeys ).join( '|' ),
```
wiese: This is not incorrect but weird. A variable local to this test containing the options and it being used both here and injected into the constructor (line 38) would make this much more self-explanatory. Think of the test as a documentation of what the code does - "variable injected here, has an effect there".

tests/unit/server/data-access/MwBotWikibaseMessagesRepo.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+			} );
+		} );
+
+		it( 'rejects in case of unkown message-key', ( done ) => {
+			const inLanguage = 'de';
+			const results = {
+				batchcomplete: '',
+				query: {
+					allmessages: [ {
+						'name': 'wikibase - edit',
```
wiese:
```suggestion
						'name': 'wikibase-edit',
```

src/components/mixins/Messages.ts
```diff
@@ -24,7 +24,7 @@ export interface MessagesMixin extends Vue {
 	},
 } )
 export default class Messages extends ( Vue as VueConstructor<MessagesMixin> ) {
-	public MESSAGE_KEYS = MessageKeys;
+	public readonly MESSAGE_KEYS = MessageKeys;
```
bitPogo: If we want it as const, it should walk like a const

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,82 @@
...
+					reject( new TechnicalProblem( reason ) );
+				} );
+		} );
+	}
+
+	private getMessageTranslationCollection(
+		inLanguage: string,
+		response: AllMessagesResponseMessage[],
+	): MessageTranslationCollection {
+		const messageCollection: MessageTranslationCollection = {};
```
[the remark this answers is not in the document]
bitPogo: done

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,82 @@
...
+		messageCollection[ inLanguage ] = this.transformMessages( response );
```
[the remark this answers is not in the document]
bitPogo: done

src/server/data-access/MwBotWikibaseMessagesRepo.ts
```diff
@@ -0,0 +1,82 @@
...
+		return messageCollection;
```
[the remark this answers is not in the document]
bitPogo: and done...

tests/unit/server/data-access/MwBotWikibaseMessagesRepo.spec.ts
bitPogo: done

tests/unit/server/data-access/MwBotWikibaseMessagesRepo.spec.ts
bitPogo: done

src/components/mixins/Messages.ts
wiese: (not part of this PR really)
We both agreed with your opinion when doing this change, the current state is a result of what is technically possible AFAIK. https://github.com/vuejs/vue/issues/6004

PR #152, opened 2019-01-07
edit pen: hide if not editable

wiese:
Wrap the edit pen(s) in a proprietary element which will be stripped
from the resulting mark-up (how depends on the user's edit permission).
This is not determined at rendering time based on a flag as it would
would have to be factored into the caching (hard in mw lifecycle).
This binds us to mediawiki/wikibase as it produces proprietary HTML
output that only wikibase as a middleware can understand/treat while
ordinary (modern) browsers will gracefully ignore it but would not know
what to make of it should it ever reach them directly.

This is probably not the most beautiful way to do it in vue but could
not find a cleaner way that actually worked. Tried
* a SSR-only directive https://ssr.vuejs.org/api/#directives
* a functional component https://vue-loader.vuejs.org/guide/functional.html, https://zendev.com/2018/05/07/multi-root-vue-components.html

This obsoletes https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/480805

Bug: T212136


bitPogo:
A side question: Why we use not the api to resovle, if the user has the permission to edit the page or not?  Like: https://www.wikidata.org/w/api.php?action=query&prop=info&inprop=protection&titles=Q231&intestactions=edit maybe in combination with https://www.wikidata.org/w/api.php?action=query&list=users&ususers=Ausername&usprop=groups

wiese:
@bitPogo Excellent question. Your suggested approach would be one way, another would be to simply pass the permission in question from mw.
Both would mean that either the username or the permission would have to become a cache dimension.
In addition, the way the mw lifecycle currently looks, at the moment when the views are rendered (that is also when our SSR is called) the user as such and derived information from it are not even available. This _could_ certainly be changed, but until we know what we'll technically be _allowed_ to do about the requests to begin with I went with the pragmatic, albeit far from purest solution to the problem.

jakobw:
Rebased. Feel free to merge!

PR #156, opened 2019-01-08
Provide helpers for building action and mutation names

jakobw:


src/store/util.ts
```diff
@@ -0,0 +1,6 @@
+function buildActionOrMutationName( namespace: string, actionName: string ): string {
```
bitPogo: This could be more generic, in reason of there could be n-namespaces. The heads are always namespaces and the tail are the action.

jakobw: Good idea! Done.

PR #158, opened 2019-01-08
debugging: log server errors to console

wiese:
Log - not otherwise explicitly handled - errors to console
(i.e. stdout) to be able to debug.
This also omits potential security problems that could arise from
sending stack traces to browsers during technical problems.
Could be worthwile to do this conditionally based on env, or even
implement some logger that facilitates a centralized logging
infrastructure for prod.


wiese:
@bitPogo Thank you. Please feel free to merge things if you approve of them - one +2 is enough.

PR #160, opened 2019-01-09
Add "In-All-Languages" link

bitPogo:


src/common/MessageKeys.ts
```diff
@@ -1,4 +1,5 @@
 export enum MessageKeys {
 	EDIT = 'wikibase-edit',
 	ALIAS_SEPARATOR = 'wikibase-aliases-separator',
+	ALL_LANG = 'wikibase-entitytermsforlanguagelistview-more',
```
wiese: Please let's not use abbreviations, it's a slippery slope.

src/components/App.vue
```diff
@@ -26,6 +28,7 @@ interface AppBindings extends Vue {
 @Component( {
 	components: {
 		TermBox,
+		InMoreLanguageSwitch,
```
wiese: Just to be completely explicit. We are building a button with the "Label for the button to show all non-empty sets of terms". So it should be called `In**All**LanguageSwitch` then. Not an awesome name either but depends how we organize its content (incl. hidden element, ...).

src/components/InMoreLanguageSwitch.vue
```diff
@@ -0,0 +1,55 @@
...
+	</div>
+</template>
+
+<script lang="ts">
+import { VueConstructor } from 'vue';
+import Component, { mixins } from 'vue-class-component';
+import Messages, { MessagesMixin } from './mixins/Messages';
+
+@Component
+export default class TermBox extends ( mixins( Messages ) as VueConstructor<MessagesMixin> ) {
```
wiese: The class name is not aligned with the file name.

src/components/InMoreLanguageSwitch.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+	<div class="wikibase-termbox-subsection-switch">
+		<div class="wikibase-termbox-subsection-switch__switch">
+			<a :href="link" @click="clientShowAllLanguages( $event )">{{ message( MESSAGE_KEYS.ALL_LANG ) }}</a>
```
wiese:
```suggestion
			<a :href="link" @click="clientShowAllLanguages">{{ message( MESSAGE_KEYS.ALL_LANG ) }}</a>
```

The event is passed automagically.

src/components/InMoreLanguageSwitch.vue
```diff
@@ -0,0 +1,55 @@
...
+export default class TermBox extends ( mixins( Messages ) as VueConstructor<MessagesMixin> ) {
+	public readonly link: string;
+	constructor() {
+		super();
+		this.link = 'place-holder-link';
+	}
+
+	public clientShowAllLanguages( event: MouseEvent ) {
+		console.log( 'TODO flip all-languages.' );
+		if ( event ) {
```
wiese: I would have hoped that using a solid framework rids us from such conditionals. Did you encounter a situation where there was no event?

src/components/InMoreLanguageSwitch.vue
```diff
@@ -0,0 +1,55 @@
...
+			event.preventDefault();
+		}
+	}
+}
+</script>
+
+<style lang="scss">
+.wikibase-termbox-subsection-switch {
```
wiese: Note to self: figure out how to model the Block part consistently.

tests/unit/components/InMoreLanguageSwitch.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+	NS_MESSAGES,
+} from '@/store/namespaces';
+import { MESSAGES_INIT } from '@/store/messages/mutationTypes';
+import { LANGUAGE_INIT } from '@/store/user/mutationTypes';
+import { mutation } from '@/store/util';
+
+const localVue = createLocalVue();
+localVue.use( Vuex );
+
+describe( 'InMoreLanguageSwitch.vue', () => {
```
wiese:
```suggestion
describe( 'InMoreLanguageSwitch', () => {
```

tests/unit/components/InMoreLanguageSwitch.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+	);
+
+	store.commit( mutation( NS_USER, LANGUAGE_INIT ), 'de' );
+
+	it( 'renders the link', () => {
+		const wrapper = shallowMount( InMoreLanguageSwitch, { store, localVue } );
+		expect( wrapper.find( '.wikibase-termbox-subsection-switch__switch>a' ).text() ).toBe( deAllShowMessages );
+	} );
+
+	it( 'renders correctly the reference', () => {
```
wiese: Yoda would love this

src/components/InMoreLanguageSwitch.vue
```diff
@@ -0,0 +1,55 @@
...
+	
+	.wikibase-termbox-subsection-switch {
+		&__switch {
+			height: 100%;
+			display: flex;
+    		align-items: center;
+			justify-content: center;
+		}
+		
+		&__switch>a {
```
wiese: Quite a bit of CSS given the story wanted an unstyled link. Is this now per design spec and the respective story (not part of the current sprint) is done through this, or some intermediate step that still requires the other story? If the latter I'm in favor of not having any style at all at this moment.

src/components/InMoreLanguageSwitch.vue
```diff
@@ -0,0 +1,55 @@
...
+</template>
+
+<script lang="ts">
+import { VueConstructor } from 'vue';
+import Component, { mixins } from 'vue-class-component';
+import Messages, { MessagesMixin } from './mixins/Messages';
+
+@Component
+export default class TermBox extends ( mixins( Messages ) as VueConstructor<MessagesMixin> ) {
+	public readonly link: string;
```
wiese: This should be solvable through component annotation
```
@Component( {
	link: {
		href: String,
		default: 'place-holder-link',
	},
} )
```

src/components/InMoreLanguageSwitch.vue
```diff
@@ -0,0 +1,55 @@
...
+	constructor() {
+		super();
+		this.link = 'place-holder-link';
+	}
+
+	public clientShowAllLanguages( event: MouseEvent ) {
```
jakobw: This might only ever be triggered in the client, but I don't that it has "client" in the name, especially since it doesn't need to get an event as an input.

tests/unit/components/InMoreLanguageSwitch.spec.ts
bitPogo: We did this also in the other components

src/common/MessageKeys.ts
bitPogo: changed

src/components/App.vue
bitPogo: changed

src/components/InMoreLanguageSwitch.vue
bitPogo: c/p error...changed

src/components/InMoreLanguageSwitch.vue
bitPogo: see below

src/components/InMoreLanguageSwitch.vue
bitPogo: this was just what was the the doc says...(more or less) see: https://vuejs.org/v2/guide/events.html#Event-Modifiers (last code block)....and it is changed

src/components/InMoreLanguageSwitch.vue
bitPogo: done

src/components/InMoreLanguageSwitch.vue
wiese: Well. To me it feels like passing the buck of testing that prop to the next guy who has to make the link injectable - as such, though generally agreeing with what you said, I feel this does not apply here.

bitPogo: From my point of view, I prefer to get the link from the store (because of that single source of truth thing)...I guess there will be a something in between of both approaches...but first of all we need the link, then we can discuss to tackle the problem, right? So it's resolved?

PR #161, opened 2019-01-09
Determine entity editability on init

jakobw:


src/components/App.vue
```diff
@@ -42,6 +42,7 @@ export default class App extends ( Vue as VueConstructor<AppBindings> ) {
 		return Promise.all( [
 			store.dispatch( action( NS_LANGUAGE, LANGUAGE_INIT ) ),
 			store.dispatch( action( NS_ENTITY, ENTITY_INIT ), request.entityId ),
+			store.dispatch( action( NS_ENTITY, DETERMINE_EDITABILITY ) ),
```
wiese: I'd be more comfortable with this being invoked as a consequence of `action( NS_ENTITY, ENTITY_INIT )` - can you relate to that or does it feels like this wires too much use case specific stuff into a store that should work for all use cases? Let's discuss.

src/common/data-access/EntityEditabilityResolver.ts
```diff
@@ -0,0 +1,5 @@
+export default interface EntityEditabilityResolver {
+
+	isEditable(): Promise<boolean>;
```
wiese: Granted it would not be of much use ATM but not passing an entity (identifier) when figuring out its editability is... interesting.

tests/unit/store/entity/mutations.spec.ts
```diff
@@ -53,4 +57,14 @@ describe( 'entity/mutations', () => {
 
 	} );
 
+	it( EDITABILITY_UPDATE, () => {
+		const store = newMinimalStore( { isEditable: true } );
```
wiese: No harm done but the default state here is the opposite of what is the case in the real store.

tests/unit/store/entity/actions.spec.ts
```diff
@@ -49,4 +51,18 @@ describe( 'entity/actions', () => {
...
+		factory.setEntityEditabilityResolver( {
+			isEditable: () => Promise.resolve( true ),
+		} );
+
+		const context = {
+			commit: jest.fn(),
+		};
+
+		return ( actions[ DETERMINE_EDITABILITY ] as any )( context ).then( () => {
+			expect( context.commit ).toHaveBeenCalledWith( EDITABILITY_UPDATE, true );
```
wiese: The pickiest of nitpicks: the `true` in a variable would make the test even more self-explanatory.

src/store/entity/index.ts
```diff
@@ -10,6 +10,7 @@ export default function (): Module<Entity, any> {
 		labels: {},
 		descriptions: {},
 		aliases: {},
+		isEditable: false,
```
wiese: Can not wait to see how we will be getting this value out of here.

src/client-entry.ts
```diff
@@ -38,6 +38,14 @@ factory.setEntityRepository( new EntityRepository(
 	( window as MwWindow ).mw.hook( Hooks.entityLoaded ),
 ) );
 
+factory.setEntityEditabilityResolver( {
+	isEditable() {
+		return Promise.resolve(
+			( window as MwWindow ).mw.config.get( 'wgIsProbablyEditable' ),
```
wiese: `wgIsProbablyEditable` is missing in `src/mock-data/MwConfig.ts`.

jakobw: done

src/common/data-access/EntityEditabilityResolver.ts
jakobw: added!

src/components/App.vue
jakobw: done

tests/unit/store/entity/actions.spec.ts
jakobw: done

tests/unit/store/entity/mutations.spec.ts
jakobw: changed!

tests/unit/store/entity/actions.spec.ts
```diff
@@ -31,6 +43,21 @@ describe( 'entity/actions', () => {
...
+			factory.setEntityEditabilityResolver( {
+				isEditable: () => Promise.resolve( isEditable ),
+			} );
+
+			const context = {
+				commit: jest.fn(),
+			};
+
+			return ( actions[ ENTITY_INIT ] as any )( context, 'Q1' ).then( () => {
+				expect( context.commit ).toHaveBeenCalledWith( EDITABILITY_UPDATE, isEditable );
```
wiese: TIL `.toHaveBeenCalledWith()` and `.toBeCalledWith()` are actual aliases. https://jestjs.io/docs/en/expect#tohavebeencalledwitharg1-arg2-
Knowing that I like to former better as it seems grammatically correct when asserting after the fact.

PR #163, opened 2019-01-10
Hide the edit pen if the entity is not editable

jakobw:


wiese:
@jakobw Poke me once you want this in #161

PR #165, opened 2019-01-11
[WIP] directionality mock data: use lib

wiese:
The mock fixture name contained a language code which confused me and as
it provided nothing more than what an already installed lib (doing that
very same job on the server side) can perform, I think we should lose
it.


wiese:
Just realized this uses a server type definition.

bitPogo:
Just a reminder -> the directionalities of wb and this package are not the same, so we should keep this in mind

wiese:
If this is something we need to keep in mind then not with regards to this change (dev mocks) but with regards to SSR (using the lib) vs the CSR (using mw goodness).

jakobw (approved):
lgtm. Is this still WIP?

jakobw:
The valid question of "Why can we use a module whose type is only defined in `/server/`?" came up. I poked around and found out the following:
* we get no console output for `src/mockup-entry.ts` because it is excluded in `tsconfig.json`
* it gets built anyway because vue-cli's `serve` does not abort on errors thrown by typescript

jakobw:
I found this out by conducting the following clever experiment:
I added these two lines to client-entry.ts
```ts
const derp: object = { foo: 'bar' };
console.log( derp.foo );
```
and got an error printed to stdout, but it got built anyway. Doing the same to mockup-entry.ts had the same effect minus the logging. Removing `mockup-entry.ts` from the `exclude` in `tsconfig.json` resulted in both entry files being treated equally.

manicki:
re-submitted to Gerrit as https://gerrit.wikimedia.org/r/#/c/wikibase/termbox/+/485154

PR #168, opened 2019-01-11
[WIP] "Term" component draft

bitPogo:
Inspired by #106 
this is just a draft - let's discuss this approach


src/components/InAllLanguages.vue
```diff
@@ -0,0 +1,92 @@
...
+		<div class="wikibase-termbox__switch">
+			<a :href="link" @click.prevent="showAllLanguagesSwitch()">{{ message( MESSAGE_KEYS.ALL_LANGUAGES ) }}</a>
+		</div>
+		<div v-if="showAllLanguages">
+			<div>
+				<div v-for="language in allLanguagesLanguageKeys">
+					<Dab :modifier="modifier" :language="language" />
+				</div>
+			</div>
+			<div class="wikibase-termbox__actions">
```
jakobw: This actions element has to go. Editing for terms that are not the user language is only going to be introduced in a much later story.

src/common/TermModifiers.ts
```diff
@@ -0,0 +1,4 @@
+export enum TermModifiers {
```
jakobw: Do you think these are going to be used in more than one place, i.e. outside of the `modifier` prop? If not, I don't think they need their own enum, especially since they're very descriptively named to begin with

src/components/Dab.vue
```diff
@@ -0,0 +1,178 @@
+<template>
+	<div 
+		class="wikibase-termbox-term" 
+		:class="modifier">
+		<div>
```
jakobw: This `div` looks like it's not needed

src/components/Dab.vue
```diff
@@ -0,0 +1,178 @@
...
+			entityLabel: 'getLabelByLanguage',
+			entityDescription: 'getDescriptionByLanguage',
+			entityAliases: 'getAliasesByLanguage',
+		} ),
+		...mapGetters( NS_LANGUAGE, {
+			getLanguageTranslation: 'getTranslationByCode',
+		} ),
+	},
+} )
+export default class Dab extends ( mixins( Messages ) as VueConstructor<DabBindings> ) {
```
jakobw: I vote for `Fingerprint` for the name of this component :)

src/store/entity/getters.ts
```diff
@@ -26,4 +26,24 @@ export const getters: GetterTree<Entity, any> = {
 	getAliasesByLanguage: ( state: Entity ) => ( languageCode: string ): Term[] | null => {
 		return state.aliases[ languageCode ] || null;
 	},
+	getAvailableLanguageKeys: ( state: Entity ): string[] => {
```
jakobw: I think this is a great use case for a getter!

Naming suggestion
`getAllEnteredLanguages`/`getAllEnteredLanguageCodes`?

src/store/entity/getters.ts
```diff
@@ -26,4 +26,24 @@ export const getters: GetterTree<Entity, any> = {
...
+			] ),
+		];
+
+		unionKeys.sort( ( key1, key2 ) => {
+			key1 = key1.toLowerCase();
+			key2 = key2.toLowerCase();
+			return ( key1 < key2 ) ? -1 : ( key1 > key2 ) ? 1 : 0;
+		} );
+
+		return unionKeys;
```
jakobw: Language codes should always be lower case already, so using the default `sort` behavior should be fine:
```suggestion
		return [
			...new Set( [
				...Object.keys( state.labels ),
				...Object.keys( state.descriptions ),
				...Object.keys( state.aliases ),
			] ),
		].sort();
```

src/components/PrimaryLanguage.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+	<div class="wikibase-termbox wikibase-termbox--primaryLanguage">
+		<div>
+			<Dab :modifier="modifier" :language="primaryLanguage"/>
+		</div>
+		<div class="wikibase-termbox__actions">
```
wiese: `.wikibase-termbox__actions` style moved into `Dab.vue` so got detached. As mentioned above, I think we should move the actions (once) into `TermBox.vue` (with the style obviously).

src/store/entity/getters.ts
wiese: Especially in the light of no written specification regarding this we should be frugal investing into custom ordering.

src/components/Dab.vue
```diff
@@ -0,0 +1,178 @@
...
+		const aliases: Term[] =  this.entityAliases( this.language );
+		if ( aliases === null ) {
+			return [];
+		} else {
+			return aliases;
+		}
+	}
+
+	get languageName(): string {
+		const name = this.getLanguageTranslation( this.language, this.primaryLanguage );
```
wiese: The `primaryLanguage` property is proudly presented (only) by the `MessagesMixin`. Had a brief discussion with @jakobw about whether or not we regard this as desirable when we first introduced the mixin. Works but I don't like it.

src/components/InAllLanguages.vue
```diff
@@ -0,0 +1,92 @@
...
+	},
+} )
+export default class InAllLanguages extends ( mixins( Messages ) as VueConstructor<InAllLanguagesBindings> ) {
+	public link = 'place-holder-link';
+	public readonly modifier = TermModifiers.ALL_LANGUAGES;
+	public allLanguagesSwitch = false;
+	public allLanguagesLanguageKeysLock = false;
+	public allLanguagesLanguageKeys: string[] = [];
+
+	get showAllLanguages(): boolean {
```
wiese: Either I missed something or this was added as a bonus. I'd love to see this as a separate commit so that we can reason about it in isolation...

bitPogo:
Done by #177

PR #172, opened 2019-01-14
Introduce getAllEnteredLanguageCodes into entity store

bitPogo:


bitPogo:
Merge the chain first

tests/unit/store/entity/getters.spec.ts
```diff
@@ -119,6 +119,25 @@ describe( 'entity/Getters', () => {
 			)( 'de' ) ).toBe( null );
 		} );
 
-	} );
+		it( 'returns all entered language-codes on a entity', () => {
```
jakobw:
```suggestion
		it( 'getAllEnteredLanguageCodes returns all entered language-codes on a entity', () => {
```

PR #174, opened 2019-01-14
Introduce Fingerprint component

bitPogo:


bitPogo:
See 176&177 how it fits into Termbox

src/components/Fingerprint.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+	<div 
+		class="wikibase-termbox-fingerprint" 
+		:class="modifier()">
```
wiese: Vue can do such mapping via
```suggestion
		:class="{ 'wikibase-termbox-fingerprint--primaryLanguage': isModified }">
```

However I did not quite get behind the naming of `isModified` - what's the story?

src/components/Fingerprint.vue
```diff
@@ -0,0 +1,113 @@
...
+		if ( name === null ) {
+			return '????';
+		} else {
+			return name;
+		}
+	}
+}
+</script>
+
+<style lang="scss">
```
wiese: For me this step in the chain of commits is the "move what we identified as the fingerprint component into its own file".
Why did the styles not move from `TermBox.vue` into this file, too?
And why was the mark-up that is now duplicated (Fingerprint.vue and `TermBox.vue`) not replaced by the use of `<Fingerprint>` inside `TermBox.vue`? That, to me, would have been one atomic change.
Reason why I think this is important is that only this way will we be able to avoid inadvertently creating integration problems that we'll have to fix down the line. 
https://pbs.twimg.com/media/Ci9dn7vWYAAGbuV.jpg

wiese:
Fixed `dist/`

src/components/Fingerprint.vue
```diff
@@ -0,0 +1,162 @@
+<template>
+	<div 
+		class="wikibase-termbox-fingerprint" 
+		:class="{ 'wikibase-termbox-fingerprint--primaryLanguage': isPrimary }">
```
wiese:
```suggestion
		:class="{ 'wikibase-termbox-fingerprint--primaryLanguage': isPrimary }"
		:dir="nowwhat?"
		:lang="language">
```

src/components/Fingerprint.vue
```diff
@@ -0,0 +1,162 @@
...
+		<span class="wikibase-termbox-fingerprint__language">{{ languageName }}</span>
```
wiese:
```suggestion
		<span class="wikibase-termbox-fingerprint__language"
			:lang="primaryLanguage"
			:dir="nowwhat?">{{ languageName }}</span>
```

src/components/Fingerprint.vue
wiese: Will create follow up.

PR #178, opened 2019-01-15
Add toggle "in all languages" behavior

jakobw:
Bug: T213532


src/components/Fingerprint.vue
```diff
@@ -104,6 +104,14 @@ export default class Fingerprint extends ( mixins( Messages ) as VueConstructor<
 <style lang="scss">
 .wikibase-termbox-fingerprint {
 
+	.wikibase-termbox & {
```
wiese:
```suggestion
	.wikibase-termbox & { // this serves as strong selector to overcome reset.css
```

src/components/TermBox.vue
```diff
@@ -36,9 +41,12 @@ export default class TermBox extends Vue {
 
 <style lang="scss">
 .wikibase-termbox { // container - need a strong selector chain to reliably override reset css
-	display: flex;
 
 	.wikibase-termbox { // for use as a prefix
+		&__primary {
+			display: flex;
```
wiese: Now we are flexing all around - not a bad thing but maybe we can specify do this once on a high level (instead of time again with using extra classes) and just add controlling features in places? Queue reading session.

tests/unit/components/AllEnteredLanguagesExpandable.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+		expect( wrapper.find( '.wikibase-termbox-subsection-switch > a' ).exists() ).toBeTruthy();
+	} );
+
+	it( 'expands all entered languages on click', () => {
+		const wrapper = shallowMount( AllEnteredLanguagesExpandable, { store } );
+		wrapper.find( '.wikibase-termbox-subsection-switch > a' ).trigger( 'click' );
+
+		expect( wrapper.find( AllEnteredLanguages ).exists() ).toBeTruthy();
+	} );
+
```
wiese: It half proven by the previous test, but nowhere do we really assert it is closed by default.
```suggestion

	it( 'does not expand all entered languages by default', () => {
		const wrapper = shallowMount( AllEnteredLanguagesExpandable, { store } );
		expect( wrapper.find( AllEnteredLanguages ).exists() ).toBeFalse();
	} );
```

src/components/Fingerprint.vue
jakobw: thanks! added

src/components/TermBox.vue
jakobw: I'm pretty sure we need this on a per-flexy-element basis. This one is here so that the edit pen sits next to it.

tests/unit/components/AllEnteredLanguagesExpandable.spec.ts
jakobw: good point. added!

src/components/TermBox.vue
wiese: Agreed. "enables a flex context for all its _direct_ children." https://css-tricks.com/snippets/css/a-guide-to-flexbox/

PR #179, opened 2019-01-15
Make non-primary fingerprint responsive.

jakobw:


src/components/Fingerprint.vue
```diff
@@ -166,5 +168,21 @@ export default class Fingerprint extends ( mixins( Messages ) as VueConstructor<
...
+			.wikibase-termbox-fingerprint__terms {
+				display: flex;
+				.wikibase-termbox-fingerprint__label,
+				.wikibase-termbox-fingerprint__description,
+				.wikibase-termbox-fingerprint__aliases {
+					flex: 1;
+				}
+				.wikibase-termbox-fingerprint__description,
+				.wikibase-termbox-fingerprint__aliases {
+					margin-left: 16px;
```
wiese: When in "table mode" `&__description` & `&__aliases` should lose their `margin-top` (otherwise used for the vertical spacing between them) to be aligned with the label.
```suggestion
					margin-left: 16px;
					margin-top: 0;
```

src/components/Fingerprint.vue
```diff
@@ -166,5 +168,21 @@ export default class Fingerprint extends ( mixins( Messages ) as VueConstructor<
 			@include fontSize( 23px );
 		}
 	}
+	.wikibase-termbox-fingerprint-list & {
+		@include media-breakpoint-up( md ) {
```
wiese: Flex - hurray.
Now to the `lg` break point (1036px+) which is supposed to show 4 columns.

src/components/Fingerprint.vue
```diff
@@ -166,5 +168,21 @@ export default class Fingerprint extends ( mixins( Messages ) as VueConstructor<
 			@include fontSize( 23px );
 		}
 	}
+	.wikibase-termbox-fingerprint-list & {
```
wiese: This is a class (only) existing in a parent component. No good.
If this is to negate the `--primaryLanguage` modifier we could
* use a second modifier and use that whenever we are not showing primary language content
* define the default, non-primary language CSS first, and explicitly override what is different for the primary language

I'm leaning towards the latter but it sure _could_ be verbose...

jakobw: Using `:not` seems fine!

src/components/Fingerprint.vue
jakobw: done

src/components/Fingerprint.vue
jakobw: done

PR #181, opened 2019-01-16
Store returns  languages in a alphabetical sorted order

bitPogo:


src/store/entity/getters.ts
```diff
@@ -27,12 +27,18 @@ export const getters: GetterTree<Entity, any> = {
 		return state.aliases[ languageCode ] || null;
 	},
 	getAllEnteredLanguageCodes: ( state: Entity ): string[] => {
-		return [
-			...new Set( [
-				...Object.keys( state.labels ),
-				...Object.keys( state.descriptions ),
-				...Object.keys( state.aliases ),
-			] ),
+		const union = [
```
jakobw: Why not just
```ts
		return [
			...new Set( [
				...Object.keys( state.labels ),
				...Object.keys( state.descriptions ),
				...Object.keys( state.aliases ),
			] ),
		].sort();
```

bitPogo: There was a reason, which I cannot reproduce anymore

PR #184, opened 2019-01-16
Enforce Termbox min-width

jakobw:
* prevents the text making the flex element wider than it should be
* factors the margin into the description/alias width

Bug: T213538


bitPogo:
Is this pen behavior intended?
penmov1
penmov2

src/components/Fingerprint.vue
```diff
@@ -119,13 +120,22 @@ export default class Fingerprint extends ( mixins( Messages ) as VueConstructor<
...
 		&__description,
 		&__aliases {
-			min-width: 260px;
+			min-width: $fingerprintMinWidth;
 			max-width: 420px;
 		}
 
+		&__description,
+		&__aliases {
+			min-width: calc(#{$fingerprintMinWidth} - #{$descriptionAliasIndent});
```
bitPogo: What is about Opera mini ( https://caniuse.com/#feat=calc )

jakobw: Well spotted, I forgot about this one.
I'm trying to solve the following problem: limit the description/alias container width to 260px while also giving it a margin-left. `calc` is one way to do it, but we can't use that. I now added a wrapping element for description and alias to achieve this. Better ideas are welcome!

jakobw:
Thanks, @bitPogo for trying it out. On the first picture everything is fine, right? The second one clearly doesn't look ideal, however I believe what you're seeing is the min-width of 260px being enforced. Also, you've set the view port width to 300px, which is way below the specified min-width of 340px. See my comment here https://phabricator.wikimedia.org/T213538#4877040

Here is what it looks like with a 340px view port:
screenshot from 2019-01-17 11-30-35

bitPogo:
I guess there no pretty way to avoid that behavior, currently...

wiese:
Discovered https://phabricator.wikimedia.org/T214037 as I went but could not prove this was the cause.

PR #185, opened 2019-01-16
Fingerprint: add language attributes in mark-up

wiese:
Bug: T213926

PR #186, opened 2019-01-16
LanguageNameInUserLanguage: mv into dedicated component

wiese:
Move translation of language names into user language into its own
component and ensure its mark-up contains relevant language attributes.

Bug: T213926


wiese:
Goes into #185

wiese:
Rebased

tests/edge-to-edge/server/app.spec.ts
```diff
@@ -297,8 +297,13 @@ describe( 'Termbox SSR', () => {
 			expect( $primaryLanguageFingerprint )
 				.toHaveAttribute( 'lang', language );
 
-			expect( $dom.querySelector( '.wikibase-termbox-fingerprint__language' ) )
+			const $language = $dom.querySelector( '.wikibase-termbox-fingerprint__language' ) as Element;
```
jakobw:
```suggestion
			const $language = $dom.querySelector( '.wikibase-termbox-fingerprint__language' );
```

src/components/Fingerprint.vue
```diff
@@ -24,17 +24,14 @@
 import Vue, { VueConstructor } from 'vue';
 import Component, { mixins } from 'vue-class-component';
 import Messages, { MessagesMixin } from './mixins/Messages';
-import {
```
jakobw, line 27: It's nice to see some code move out of here!

jakobw (commented):
Cool!
screenshot from 2019-01-17 15-40-56

tests/edge-to-edge/server/app.spec.ts
wiese: Probably goes without saying but there were $elements which were further queried down the line, causing compiler errors because the previous query can yield null results and assertions do not act as sufficient guards as far as tsc is concerned.

jakobw (approved):

PR #189, opened 2019-01-17
[WIP] Prevent unwanted wrapping of flex elements

jakobw:
I tried to describe the cause for the behavior here: https://phabricator.wikimedia.org/T214037#4888362

I'm not so happy about adding yet another wrapping element, but I like the flex/no-flex of the language better this way than with the `flex-basis: 100%` hack. Looks like this at 892px:

screenshot from 2019-01-17 16-52-36


wiese:
We appear to still be struggling in case of label, description, alias vastly differing in width (because of amount of content).
bildschirmfoto von 2019-01-17 18-47-17

manicki:
re-submitting to Gerrit as https://gerrit.wikimedia.org/r/#/c/wikibase/termbox/+/485155

